PR #30, opened 2020-02-20
Queue new API + docs

nomisRev:
This PR exposes the new APIs from Queue, peek and try that do not block.
It also includes additional docs and tests for the new APIs, and a small restructuring.

It seems I have incorrectly merged ConcurrentQueue with CancelableQueue tho.
Any1 any idea how I keep the git history correctly? I'd happily re-open a new PR.


aballano:
> Any1 any idea how I keep the git history correctly? I'd happily re-open a new PR.

If you mean within this PR it's already done when you squash, just edit the message from here. If you mean from the master history then I don't think so because we would need to rewrite history by force-pushing, which would mean anyone with the branch might have conflicts.

nomisRev:
@aballano that's not what I meant.
It looks like I renamed `ConcurrentQueue` to `CancellableQueue`, but I removed `CancellableQueue` and moved/refactored `ConcurrentQueue`.

Not sure how I can reflect that change properly in git. I retried with intermediate commits but got the same history.

Not critical, so ready for review. :)

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,232 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+  /**
+   * Take a value from the [Queue], or sementically blocks until a value becomes available.
+   *
+   * @see [peek] for a function that doesn't remove the value from the [Queue].
+   * @see [tryTake] for a function that does not sementically block but returns immediately with an [Option].
+   */
   fun take(): Kind<F, A>
+
+  /**
+   * Tries to take a value from the [Queue]. Returns immediately with either [None] or a value [Just].
```
1Jajen1:
```suggestion
   * Tries to take a value from the [Queue]. Returns immediately with either [None] or a value [Some].
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,232 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
 inline fun <F, A> QueueOf<F, A>.fix(): Queue<F, A> =
   this as Queue<F, A>
 
+/** A polymoprhic effect typeclass that allows [Dequeue]'ing values from a [Queue]. */
 interface Dequeue<F, A> {
+
+  /**
+   * Take a value from the [Queue], or sementically blocks until a value becomes available.
```
1Jajen1:
```suggestion
   * Take a value from the [Queue], or semantically blocks until a value becomes available.
```

This occurs in quite a few places.

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,232 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+  /**
+   * Peeks a value from the [Queue] or semantically blocks until one becomes available.
+   * In contrast to [take], [peek] does not remove the value from the [Queue].
+   *
+   * @see [tryPeek] for a function that does not sementically blocks but returns immediately with an [Option].
+   */
+  fun peek(): Kind<F, A>
+
+  /**
+   * Tries to peek a value from the [Queue]. Returns immediately with either [None] or a value [Just].
```
1Jajen1:
```suggestion
   * Tries to peek a value from the [Queue]. Returns immediately with either [None] or a value [Some].
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,232 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+   * In contrast to [tryTake], [tryPeek] does not remove the value from the [Queue].
+   *
+   * @see [peek] for a function that sementically blocks until a value becomes available.
+   */
+  fun tryPeek(): Kind<F, Option<A>>
 }
 
+/** A polymoprhic effect typeclass that allows [Dequeue]'ing values from a [Queue]. */
```
1Jajen1:
```suggestion
/** A polymorphic effect typeclass that allows [Enqueue]'ing values from a [Queue]. */
```

1Jajen1:
Oh btw is it possible to support bulk operations like `flush` or `offerAll`? Those could be useful if you don't want to start fibers for each value to add/remove.

nomisRev:
Yes, we definitely can but I haven't attempted to implement them yet.

nomisRev:
@1Jajen1 what behavior would you expect for the different backpressure strategies? I think the one described below would be quite easy to add.

## offerAll(la: Iterable<A>): Kind<F, Unit>

- Unbounded#offerAll: always adds all immediately
- Bounded#offerAll: blocks until all are added, adds every element as soon as soon as capacity available. This way you can add more than the capacity size.
- Dropping#offerAll: adds all in order, respecting capacity
- Sliding#offerAll: adds all in order, respecting capacity

## flushAll(): Kind<F, List<A>>

For all strategies returns what is currently in `Queue`

1Jajen1:
I'd expect it to work as if we'd repeat `offer` so that: `offerAll(xs) == traverse_ { a -> offer(a) }` or similar, which includes all the blocking/dropping etc.

Guessing from how the strategies currently work what you suggested should be the same right?

The same cannot really hold for `flush` it would be more like a recursive `tryTake` until it reaches `None`, so `flush` should never block.

nomisRev:
Right, that's exactly what I was thinking.

So no need to implement anything low level, I'll add these as derived methods to `Queue` with tests to verify what I mentioned above.

1Jajen1:
> Right, that's exactly what I was thinking.
> 
> So no need to implement anything low level, I'll add these as derived methods to `Queue` with tests to verify what I mentioned above.

That depends, are these methods supposed to be atomic? Or better put does this hold: `offerAll(xs).fork().followedBy(offer(x).fork()) => Queue(x + xs) or Queue(xs + x)`?

Your name suggestions (missed them the first time) also make sense: `takeAll` `peekAll` both work for me.

nomisRev:
`offerAll(xs).fork().followedBy(offer(x).fork())` this can never be atomic, no matter how it's implemented. two async forks and a bind.

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
 inline fun <F, A> QueueOf<F, A>.fix(): Queue<F, A> =
   this as Queue<F, A>
 
+/** A polymoprhic effect typeclass that allows [Dequeue]'ing values from a [Queue]. */
```
aballano:
```suggestion
/** A polymorphic effect typeclass that allows [Dequeue]'ing values from a [Queue]. */
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
 interface Dequeue<F, A> {
+
+  /**
+   * Take a value from the [Queue], or sementically blocks until a value becomes available.
+   *
+   * @see [peek] for a function that doesn't remove the value from the [Queue].
+   * @see [tryTake] for a function that does not sementically block but returns immediately with an [Option].
```
aballano:
```suggestion
   * @see [tryTake] for a function that does not semantically block but returns immediately with an [Option].
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+   */
   fun take(): Kind<F, A>
+
+  /**
+   * Tries to take a value from the [Queue]. Returns immediately with either [None] or a value [Just].
+   *
+   * @see [take] for function that sementically blocks until a value becomes available.
+   * @see [tryPeek] for a function that sementically blocks until a value becomes available.
```
aballano:
```suggestion
   * @see [take] for function that semantically blocks until a value becomes available.
   * @see [tryPeek] for a function that semantically blocks until a value becomes available.
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+   */
+  fun tryTake(): Kind<F, Option<A>>
+
+  /**
+   * Peeks a value from the [Queue] or semantically blocks until one becomes available.
+   * In contrast to [take], [peek] does not remove the value from the [Queue].
+   *
+   * @see [tryPeek] for a function that does not sementically blocks but returns immediately with an [Option].
```
aballano:
```suggestion
   * @see [tryPeek] for a function that does not semantically blocks but returns immediately with an [Option].
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+   */
+  fun peek(): Kind<F, A>
+
+  /**
+   * Tries to peek a value from the [Queue]. Returns immediately with either [None] or a value [Just].
+   * In contrast to [tryTake], [tryPeek] does not remove the value from the [Queue].
+   *
+   * @see [peek] for a function that sementically blocks until a value becomes available.
```
aballano:
```suggestion
   * @see [peek] for a function that semantically blocks until a value becomes available.
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+  fun peekAll(): Kind<F, List<A>>
 }
 
+/** A polymoprhic effect typeclass that allows [Dequeue]'ing values from a [Queue]. */
 interface Enqueue<F, A> {
+
+  /**
+   * Offers a value to the [Queue], and might behave differently depending on the [Queue.BackpressureStrategy].
+   *
+   *  - Sementically blocks until room available in [Queue] for [Queue.BackpressureStrategy.Bounded]
```
aballano:
```suggestion
   *  - Semantically blocks until room available in [Queue] for [Queue.BackpressureStrategy.Bounded]
```

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+   *  - Returns immediately and slides values through the [Queue] for [Queue.BackpressureStrategy.Sliding]
```
aballano: Not sure I understand the meaning of "slide" when talking about offering, maybe an explanation like https://github.com/arrow-kt/arrow-fx/pull/30/files#diff-f560b92378f730cad2c0c8916a5e0071R105 would be more descriptive?

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,244 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
-   * Deficit:
-   *  Contains a queue of values and a queue of suspended fibers
-   *  waiting to take once a value becomes available
-   * Surplus:
-   *  Contains a queue of values and a queue of suspended fibers
-   *  waiting to offer once there is room (if the queue is bounded)
-   * Shutdown:
-   *  Holds no values or promises for suspended calls,
-   *  an offer or take in Shutdown state creates a QueueShutdown error
+   * Sementically blocks until the [Queue] is [shutdown].
```
aballano:
```suggestion
   * Semantically blocks until the [Queue] is [shutdown].
```

arrow-fx/src/main/kotlin/arrow/fx/internal/ConcurrentQueue.kt
```diff
@@ -78,32 +36,54 @@ class CancellableQueue<F, A> internal constructor(
...
+    }
 
-  fun read(): Kind<F, A> =
-    cancellable(::unsafeRead)
+  override fun peekAll(): Kind<F, List<A>> =
+    defer { unsafePeekAll() }
+
+  /**
+   * Waits until the queue is shutdown.
+   * The `IO` returned by this method will not resume until the queue has been shutdown.
```
aballano: Are we assuming CF = IO here?

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -20,95 +19,254 @@ typealias QueuePartialOf<F> = Kind<ForQueue, F>
...
+   * @see [peek] for a function that doesn't remove the value from the [Queue].
+   * @see [tryTake] for a function that does not sementically block but returns immediately with an [Option].
+   */
   fun take(): Kind<F, A>
+
+  /**
+   * Tries to take a value from the [Queue]. Returns immediately with either [None] or a value [Just].
+   *
+   * @see [take] for function that sementically blocks until a value becomes available.
+   * @see [tryPeek] for a function that sementically blocks until a value becomes available.
```
1Jajen1: `tryPeek` never blocks right? So the description does not match

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -40,7 +44,67 @@ class QueueTest : UnitSpec() {
...
+          }.equalUnderTheLaw(IO.just(emptyList()), EQ())
+        }
+      }
+
+      "$label - empty queue peekAll is empty" {
+        forAll(Gen.positiveIntegers()) { n ->
+          IO.fx {
+            val q = !queue(n)
+            !q.peekAll()
+          }.equalUnderTheLaw(IO.just(emptyList()), EQ())
```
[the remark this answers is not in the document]
1Jajen1: Yea I know. The queues for the tests above aren't though and those could be combined, but only if there is some validation that this case is tested, which kotlintest cannot easily do.

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -107,7 +171,72 @@ class QueueTest : UnitSpec() {
...
+        forAll(Gen.int()) { i ->
+          IO.fx {
+            val q = !queue(10)
+            val size1 = !q.size()
+            !q.offer(i)
+            val size2 = !q.size()
+            val peeked = !q.peek()
+            val size3 = !q.size()
+            !effect { Tuple4(size1, size2, peeked, size3) shouldBe Tuple4(0, 1, i, 1) }
+          }.equalUnderTheLaw(IO.unit, EQ())
```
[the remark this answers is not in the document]
1Jajen1: Yes of course, but after the last size check the element in the queue may or may not be the one peek returned. Logically it obviously is, it is better to have that tested though. Instead of checking the size after `peek`, why not use `takeAll` and verify that it is equal to `listOf(i)`?

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -159,7 +288,82 @@ class QueueTest : UnitSpec() {
...
+          !effect { peeked shouldBe None }
+        }.equalUnderTheLaw(IO.unit, EQ())
+      }
+
+      "$label - tryTake on an empty Queue returns false" {
+        IO.fx {
+          val q = !queue(10)
+          val took = !q.tryTake()
+          !effect { took shouldBe None }
+        }.equalUnderTheLaw(IO.unit, EQ())
```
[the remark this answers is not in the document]
1Jajen1: Ah ok :+1: Did not know the counterexamples were that bad...

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -181,7 +385,161 @@ class QueueTest : UnitSpec() {
...
+            !l.parTraverse(NonEmptyList.traverse(), q::offer).fork()
+            !IO.sleep(50.milliseconds) // Give take callbacks a chance to register
+
+            val l = !q.takeAll()
+            l.toSet()
+          }.equalUnderTheLaw(IO.just(l.toList().toSet()))
+        }
+      }
+
+      "$label - peekAll takes outstanding  offers" {
```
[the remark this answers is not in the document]
1Jajen1: Makes sense, but does `peekAll` complete the offering tasks or do they still suspend? If they suspend then this is perfectly fine^^

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -181,7 +385,161 @@ class QueueTest : UnitSpec() {
...
+      "$label - takeAll takes outstanding  offers" {
+        forAll(50, Gen.nonEmptyList(Gen.int()).filter { it.size <= 100 }, Gen.choose(1, 100)) { l, capacity ->
+          IO.fx {
+            val q = !queue(capacity)
+            !l.parTraverse(NonEmptyList.traverse(), q::offer).fork()
+            !IO.sleep(50.milliseconds) // Give take callbacks a chance to register
+
+            val l = !q.takeAll()
+            l.toSet()
+          }.equalUnderTheLaw(IO.just(l.toList().toSet()))
```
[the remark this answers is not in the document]
1Jajen1: Yeah my bad, mixed that up^^ The point of this should be to always have some outstanding offers which means `capacity < size`

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -246,6 +604,58 @@ class QueueTest : UnitSpec() {
...
+      "$label - tryOfferAll under capacity" {
+        forAll(Gen.list(Gen.int()).filter { it.size <= 100 }, Gen.int().filter { it > 100 }) { l, capacity ->
+          IO.fx {
+            val q = !queue(capacity)
+            val succeed = !q.tryOfferAll(l)
+            val all = !q.takeAll()
+            Tuple2(succeed, all)
+          }.equalUnderTheLaw(IO.just(Tuple2(true, l)))
+        }
+      }
```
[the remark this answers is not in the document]
1Jajen1: Blind me must have been writing this ^^ I mixed up the max/min bounds here

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
nomisRev: I'm going to improve on this before merging, I'll poke you for another tech content review.

arrow-fx/src/main/kotlin/arrow/fx/internal/ConcurrentQueue.kt
nomisRev: Internal docs. I'll update.

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -40,7 +44,67 @@ class QueueTest : UnitSpec() {
...
+      "$label - queue can be filled at once over capacity with takers" {
+        forAll(Gen.nonEmptyList(Gen.int())) { l ->
+          IO.fx {
+            val q = !queue(l.size)
+            val succeed = !q.tryOfferAll(l.toList())
+            val res = !q.takeAll()
+            Tuple2(succeed, res)
+          }.equalUnderTheLaw(IO.just(Tuple2(true, l.toList())))
+        }
+      }
```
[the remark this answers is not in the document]
nomisRev: This is not the case, I have updated the test.

The first one tests at capacity,
and the other one tests at capacity with 1 taker which allows it to offer (capacity + 1) elements to `tryOfferAll`.

I updated the tests to reflect this better, and verify that the scheduled `take` receives the first element.

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -40,7 +44,67 @@ class QueueTest : UnitSpec() {
...
+
+      "$label - takeAll takes all values from a Queue" {
+        forAll(Gen.nonEmptyList(Gen.int())) { l ->
+          IO.fx {
+            val q = !queue(l.size)
+            !l.traverse(IO.applicative(), q::offer)
+            !q.takeAll()
+          }.equalUnderTheLaw(IO.just(l.toList()), EQ())
+        }
```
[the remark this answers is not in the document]
nomisRev: Updated to check the `Queue` is empty after the first `takeAll`.

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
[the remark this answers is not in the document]
nomisRev: I'm not sure I understand your comment correctly.

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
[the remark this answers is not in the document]
nomisRev: Good point, update the test. To first `offer`, `peek` and `takeAll` and checks the product of `peek` and `takeAll`.

arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -67,13 +67,13 @@ interface Dequeue<F, A> {
   fun peekAll(): Kind<F, List<A>>
 }
 
-/** A polymoprhic effect typeclass that allows [Dequeue]'ing values from a [Queue]. */
+/** A polymorphic effect typeclass that allows [Enqueue]'ing values from a [Queue]. */
```
1Jajen1:
```suggestion
/** A polymorphic effect typeclass that allows [Enqueue]'ing values to a [Queue]. */
```
Missed that on the first read through

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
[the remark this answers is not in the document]
nomisRev: They still suspend since `peekAll` does not remove any items from the `Queue`.

It's alternative `takeAll` also completes the offering tasks, making the `Queue` effectively empty.

arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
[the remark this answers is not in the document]
nomisRev: I updated the generators to be the following.

```
Gen.nonEmptyList(Gen.int()).filter { it.size in 51..100 }, // elements to `offer` to `Queue`.
Gen.choose(1, 50) // Capacity
```

nomisRev:
@1Jajen1 after some discussion with @raulraja I decided we should not expose `shutdown` at this point.

It's not a pattern we follow in the other concurrency primitives, nor is it a pattern that we actively support. Which is exceptions for control flow, in this case cancelation flow.

On top of that Arrow Fx offers automatic cancelation support, and powerful combinators to prevent cancelation breaking and leaking to occur. So we should stay close to that.

Additionally, we can always come back to this and add `shutdown`. The other way around is more difficult.

aballano:
Got a test failing before, might be flaky

```
 arrow.fx.QueueTest > BoundedQueue - takeAll returns emptyList with waiting suspended takers FAILED
     arrow.fx.internal.TimeoutException: Duration(amount=5, timeUnit=SECONDS)
```

arrow-fx/src/main/kotlin/arrow/fx/internal/ConcurrentQueue.kt
```diff
@@ -0,0 +1,512 @@
...
+      }
+
+      true
+    }
+
+  private fun callTakeAndPeeks(a: A, take: Callback<A>?, peeks: Iterable<Callback<A>>): Kind<F, Boolean> {
+    val value = Right(a)
+    return peeks.callAll(value).flatMap {
+      if (take != null) later { take(value) }.map { true }
+      else just(true)
```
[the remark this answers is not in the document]
nomisRev: I don't find `?:` clearer than `if-else` expression here.

arrow-fx/src/main/kotlin/arrow/fx/internal/ConcurrentQueue.kt
```diff
@@ -0,0 +1,512 @@
...
+
+  /**
+   * A Queue can be in three states
+   * [Deficit]: Contains three maps of registered id & take/reads/shutdown callbacks waiting
+   *   for a value to become available.
+   *
+   * [Surplus]: Contains a queue of values and two maps of registered id & offer/shutdown callbacks waiting to
+   *   offer once there is room (if the queue is bounded, dropping or sliding).
+   *
+   * [Shutdown]: Holds no values, an offer or take in Shutdown state creates a QueueShutdown error.
```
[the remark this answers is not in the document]
nomisRev: Removed `shutdown` support, thanks for noticing!

arrow-fx-test/src/main/kotlin/arrow/fx/test/laws/Law.kt
```diff
@@ -14,6 +22,18 @@ import io.kotlintest.shouldNot
 fun <A> A.equalUnderTheLaw(b: A, eq: Eq<A>): Boolean =
   shouldBeEq(b, eq).let { true }
 
+fun <A> IOOf<A>.equalUnderTheLaw(b: IOOf<A>, EQA: Eq<A> = Eq.any(), timeout: Duration = 5.seconds): Boolean =
```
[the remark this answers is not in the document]
nomisRev: Yes, it's for a proper assertion message. The one implement through `IO.toString` is not very meaningful.

https://github.com/arrow-kt/arrow-fx/pull/30#discussion_r390170250

PR #33, opened 2020-02-20
Added Reaktive module

arkivanov:
Continuation of the closed PR in the old repo: https://github.com/arrow-kt/arrow/pull/2061

1. I found and fixed a few bugs in my implementation of the `racePair` and `raceTripple` operators. The related tests look fine now.
2. The following tests are still flaky: `acquireBracketIsNotCancelable`, `releaseBracketIsNotCancelable` (PR submitted here https://github.com/arrow-kt/arrow-test/pull/10)
3. I removed disabling `reaktiveUncaughtErrorHandler` and it does not change anything, all tests are green (except those two flaky, see point 2)


nomisRev:
> 3. I removed disabling reaktiveUncaughtErrorHandler and it does not change anything, all tests are green (except those two flaky, see point 2)

So this is different from RxJava? In the current RxJava code we reach this point in [RxJavaPlugins](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java#L382) and we can see the test logs getting spammed with `UndeliverableException`, we were unable to perform a check first to prevent sending the error for an already closed `Emitter`.

This doesn't affect the tests or behavior since the behavior is correct.

arrow-fx-reaktive/src/main/kotlin/arrow/fx/reaktive/MaybeK.kt
```diff
@@ -0,0 +1,330 @@
...
+   *   )
+   *   //sampleEnd
+   *   println_atomic(safeComputation)
+   * }
+   *  ```
+   */
+
+  fun <B> bracketCase(use: (A) -> MaybeKOf<B>, release: (A, ExitCase<Throwable>) -> MaybeKOf<Unit>): MaybeK<B> =
+    value()
+      .flatMap { value: A ->
```
nomisRev: This is not a correct bracket implementation. I just noticed that the documentation can use some additional clarification here... The idea behind `bracket` is to compose `Maybe` (or `F`), in a resource-safe way
So compose an `acquire: Maybe<A>`, `use: (A) -> Maybe<B>` and `release: (A, ExitCase) -> Maybe<Unit>` in a resource-safe way.

This is only possible, if we can  _guarantee_ that `release` runs _uncancelable_ after`acquire`.

In the snippet above you can cancel between `value()` (acquire) and `use/release` due to the `flatMap` boundary, and thus there is no guarantee that `release` will be called upon cancellation.

Additionally, we also have to make `acquire` uncancelable or the resource could still be leaked with for example `Maybe.fromCallable { resource }.flatMap { println("acquired") }`.

So `bracketCase` runs in two steps.

1. Runs `acquire` in an uncancelable manner
2. Runs `use` in a cancelable manner, while guarantee-ing `release` to be called.

You can also see that reflected in the tests mentioned in your PR in [arrow-test](https://github.com/arrow-kt/arrow-test/pull/10)

- `acquireBracketIsNotCancelable` tests that the `acquire` step cannot be cancelled.
- `releaseBracketIsNotCancelable` tests that `release` cannot be cancelled.
- `cancelOnBracketReleases` tests that `release` is called on `cancel` of use or `acquire`.
   => Additional note on last law, if you cancel during `acquire` you should immediately call `release` after `acquire` and skip `use`. That's the behavior of immediately cancelling a resource-safe program.

nomisRev: Some additional information in https://github.com/arrow-kt/arrow-test/pull/10

arkivanov:
Ok guys:

1. I applied the fix for `releaseBracketIsNotCancelable` test suggested [here](https://github.com/arrow-kt/arrow-test/pull/10#discussion_r382472311), the tests is green
2. The suggested [here](https://github.com/arrow-kt/arrow-test/pull/10#discussion_r382478348) fix for `acquireBracketIsNotCancelable` test does not work for Reaktive, I'm not sure why. Fix proposed by me in this PR works, but you say it is incorrect. So let's discuss how to fix it properly.
3. I fixed all compilation issues after pulling master
4. I have updated `AsyncLaws.derivedDefer` test since it was checking for thread name instead of thread id. Thread name might be changed by coroutines dispatches, e.g. it may append current coroutine name to the thread name.

arkivanov:
Also the task `:arrow-docs:runAnk` fails here but passes locally, not sure know how to fix.

arkivanov:
Plus I see that `parMapCancelCancelsBoth` test fails here but also passes locally. Could you please check its implementation (the test itself), maybe there are some issues as well?

aballano:
Hi @arkivanov, sorry for the delay, we've been a bit busy with the release 😅 Thank you for all the work so far in this PR, I'll try to book some time next week to take a closer look at it and see why Ank and the tests are failing on CI 👍

rachelcarmena:
Hi @arkivanov, thanks for the contribution!! Please, could you update the branch? It seems it's not possible from the GUI. Thanks!!

arkivanov:
> Hi @arkivanov, thanks for the contribution!! Please, could you update the branch? It seems it's not possible from the GUI. Thanks!!

Done. Also fixed tests compilation.

nomisRev:
Hi @arkivanov, why did you close the PR?
Time constraints? Or anything I can help with?

arkivanov:
> Hi @arkivanov, why did you close the PR?
> Time constraints? Or anything I can help with?

Hi @nomisRev,
The reason is that raised issues with flaky tests remain unaddressed for months. Now I have strict time constraints and will not be able to spend more time on this PR.
Apologize for any inconvenience.

pakoito:
With covid it's being an interesting year so there's less time to dedicate to opensource. I understand why you'd have time constraints.

I hope we get to fixing those and we have this project back in the future :D

nomisRev:
Hi @arkivanov,

Indeed strange times!
I remember you made some interesting progress concerning `(CoroutineContext) -> Scheduler`?
Is there any other changes/fixes that's worth cherry-picking or porting to the current master?

Let's try not to go any effort to go to waste :) So hopefully an integration later will be easier.

arkivanov:
> Hi @arkivanov,
> 
> Indeed strange times!
> I remember you made some interesting progress concerning `(CoroutineContext) -> Scheduler`?
> Is there any other changes/fixes that's worth cherry-picking or porting to the current master?
> 
> Let's try not to go any effort to go to waste :) So hopefully an integration later will be easier.

Hey! Yes I implemented CoroutineContext <---> Scheduler interop in the corresponding Reaktive module: https://github.com/badoo/Reaktive/tree/master/coroutines-interop/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/coroutinesinterop

Probably you can copy-paste this for RxJava and maybe for Reactor.

PR #44, opened 2020-02-25
Increase test suite for BracketLaws

### LINKED ISSUE #10: Improve polymorphic test suite
Currently the laws are not exhaustive on derived functionality. This can be breaking due to the fact that we can override derived implementations in an `interface`, which is something that is not possible for global derived functions like in Haskell.

`void :: Functor f => f a -> f ()` ~ `fun <F, A> Kind<F, A>.unit(FF: Functor<F>): Kind<F, Unit>`
Which is not override able, and this cannot be broken.

However since in Arrow they are derived within the interface, they can be overridden by more optimized concrete implementations. Which is desirable, but our test suite should make sure we keep the same guarantees when doing so.

- [x] Functor
- [x] Apply
- [x] Applicative
- [x] ApplicativeError
- [x] Monad
- [x] MonadError
- [x] MonadThrow
- [x] MonadDefer
- [ ] Bracket
- [x] Async
- [x] Concurrent
- [x] Timer
- [x] Semigroup
- [x] Monoid
- [x] Semiring

aballano:
Finishes and fixes #10


arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +168,62 @@ object BracketLaws {
...
+        .map { msg.value }
+        .equalUnderTheLaw(just(i), EQ)
+    }
+
+  fun <F> Bracket<F, Throwable>.bracketShouldBeStackSafe(
+    iterations: Int,
+    EQ: Eq<Kind<F, Int>>
+  ) {
+
+    fun ioBracketLoop(i: Int): Kind<F, Int> =
```
nomisRev:
```suggestion
    fun bracketLoop(i: Int): Kind<F, Int> =
```

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +168,62 @@ object BracketLaws {
...
+      unit().bracket(use = { just(i + 1) }, release = { unit() }).flatMap { ii ->
+        if (ii < iterations) ioBracketLoop(ii)
+        else just(ii)
+      }
+
+    forFew(100, Gen.int().applicativeError(this)) { fa ->
```
nomisRev: We run this over `forAll(Gen.create { Unit }) {` in `MonadDeferLaws.kt` and we do not run them in `forAll` in `ConcurrentLaws.kt`.

I think `forAll(Gen.create { Unit })` does a single iteration due to `Unit`.
These tests are quite slow, so running them 100x will take a considerable amount of time per datatype.

nomisRev: `Gen.int().applicativeError(this)` is not meaning-full for testing stack safety since all errors will short-circuit the chain and by-pass the stack-safety issue.

We should probably replace all these instances of `applicativeError` with `GenK`
Maybe @1Jajen1 can give some more insights on this?

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +168,62 @@ object BracketLaws {
...
+    EQ: Eq<Kind<F, Int>>
+  ) {
+    fun ioGuaranteeCase(i: Int): Kind<F, Int> =
+      unit().guaranteeCase { unit() }.flatMap {
+        val ii = i + 1
+        if (ii < iterations) ioGuaranteeCase(ii)
+        else just(ii)
+      }
+
+    forFew(100, Gen.int().applicativeError(this)) { fa ->
```
nomisRev: Idem `forFew`

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +168,62 @@ object BracketLaws {
...
+  }
+
+  fun <F> Bracket<F, Throwable>.guaranteeCaseShouldBeStackSafe(
+    iterations: Int,
+    EQ: Eq<Kind<F, Int>>
+  ) {
+    fun ioGuaranteeCase(i: Int): Kind<F, Int> =
+      unit().guaranteeCase { unit() }.flatMap {
+        val ii = i + 1
+        if (ii < iterations) ioGuaranteeCase(ii)
```
nomisRev:
```suggestion
        if (ii < iterations) guaranteeCaseLoop(ii)
```

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +168,62 @@ object BracketLaws {
...
+    forFew(100, Gen.int().applicativeError(this)) { fa ->
+      fa.flatMap { ioBracketLoop(0) }.equalUnderTheLaw(just(iterations), EQ)
+    }
+  }
+
+  fun <F> Bracket<F, Throwable>.guaranteeCaseShouldBeStackSafe(
+    iterations: Int,
+    EQ: Eq<Kind<F, Int>>
+  ) {
+    fun ioGuaranteeCase(i: Int): Kind<F, Int> =
```
nomisRev:
```suggestion
    fun guaranteeCaseLoop(i: Int): Kind<F, Int> =
```

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +168,62 @@ object BracketLaws {
...
+      unit().guaranteeCase { unit() }.flatMap {
+        val ii = i + 1
+        if (ii < iterations) ioGuaranteeCase(ii)
+        else just(ii)
+      }
+
+    forFew(100, Gen.int().applicativeError(this)) { fa ->
+      fa.flatMap { ioGuaranteeCase(0) }.equalUnderTheLaw(just(iterations), EQ)
```
nomisRev:
```suggestion
      fa.flatMap { guaranteeCaseLoop(0) }.equalUnderTheLaw(just(iterations), EQ)
```

arrow-fx-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -119,7 +119,7 @@ object ConcurrentLaws {
         Law("Concurrent Laws: RaceN arity-8 should be stack safe") { CF.race8StackSafe(iterations, EQK.liftEq(Int.eq()), ctx) },
         Law("Concurrent Laws: RaceN arity-9 should be stack safe") { CF.race9StackSafe(iterations, EQK.liftEq(Int.eq()), ctx) }
       )
-    } else emptyList()) + TimerLaws.laws(CF, CF.timer(), EQK)
```
nomisRev: Why is this removed?

aballano: Duplicated below on the public laws function

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
1Jajen1: > `Gen.int().applicativeError(this)` is not meaning-full for testing stack safety since all errors will short-circuit the chain and by-pass the stack-safety issue.
> 
> We should probably replace all these instances of `applicativeError` with `GenK`
> Maybe @1Jajen1 can give some more insights on this?

`MonadError` has laws which say `raiseError(e).flatMap(f) == raiseError(e)` so if `Gen.applicativeError` would only generate errors it'd not do anything. But `applicativeError` is defined as `Gen.oneOf` which does both states. It's still a bad gen, but it should generate both states. This test should however fail as soon as an error is generated!

1Jajen1: > I think forAll(Gen.create { Unit }) does a single iteration due to Unit.

Is that really the case? o.O I mean I know kotlintest uses sequences internally to model the random values, so this is possible, but that's weird nevertheless...

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
aballano: Done in https://github.com/arrow-kt/arrow-fx/pull/44/commits/f90752b3455b61b9bcb439a2d18cbb0fd43d7f1c

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
nomisRev: I haven't verified that statement @1Jajen1! I simply made that assumption based on the speed of the tests.

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -21,44 +22,57 @@ object BracketLaws {
...
-    )
+      Law("Bracket: guaranteeCase must run finalizer task") { BF.guaranteeCaseMustRunFinalizerOnError(EQ) },
+      Law("Bracket: bracket is derived from bracketCase") { BF.bracketIsDerivedFromBracketCase(EQ) },
+      Law("Bracket: guarantee is derived from bracket") { BF.guaranteeIsDerivedFromBracket(BF.just(Unit), EQ) },
+      Law("Bracket: guaranteeCase is derived from bracketCase") { BF.guaranteeCaseIsDerivedFromBracketCase({ BF.just(Unit) }, EQ) },
+      // onCancel cannot be tested as Bracket doesn't have the power to cancel
+      Law("Bracket: onError must run finalizer task") { BF.onErrorMustRunFinalizerOnError(EQ) }
+    ) + (if (testStackSafety) {
+      listOf(
+        Law("Bracket: onError must run finalizer task") { BF.bracketShouldBeStackSafe(iterations, EQ) },
```
nomisRev:
```suggestion
        Law("Bracket: bracket should be stack-safe") { BF.bracketShouldBeStackSafe(iterations, EQ) },
```

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -21,44 +22,57 @@ object BracketLaws {
...
+        Law("Bracket: onError must run finalizer task") { BF.guaranteeCaseShouldBeStackSafe(iterations, EQ) }
```
nomisRev:
```suggestion
        Law("Bracket: guaranteeCase should be stack-safe") { BF.guaranteeCaseShouldBeStackSafe(iterations, EQ) }
```

arrow-fx-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
1Jajen1: The assumption makes sens if they implemented `Gen.create { x } == Gen(sequenceOf(x))` or similar. Either way I'd say this is quite weird if one specifically wants to test x-times, so we should definitly test that somehow...

PR #46, opened 2020-02-25
Fix several tests relying on no-op assertion

aballano:
**BREAKING**: Make `equalUnderTheLaw` throw an exception when the evaluation is false. With this we want to make sure that all tests are **verifying something**, otherwise non-boolean tests (all tests not using `forAll/Few`) can naively call `equalUnderTheLaw` at the end but fail to remember to verify the condition as the name doesn't suggest it. This therefore caused different kind of failures:
  * Some tests failed when trying to assert over a false `equalUnderTheLaw`, either with a boolean condition or with `shouldBe` -> Simply fixed them by using a more appropriate assertion
  * [Some others](https://github.com/arrow-kt/arrow-fx/pull/46#issuecomment-591421078) failed either due to a problem with how the test was asserting or due to a **bug in the implementation**. 

All changes are explained below:

* Changed `equalUnderTheLaw` to rely on `shouldBeEq` and throw on failure
* Update Mvar read and take docs to explicitly detail if the value is removed or not
* Fix EqTest double verification causing it to fail
* Fix semaphore test wrongly calling acquire
* Fix weird stack overflow due to toString implementation of IO. This is likely caused by the `shouldBeEq` impl as it tries to print the values after the assertion and it causes a stack overflow due to the chained stack.
* Fix FreeC tests failing due to data classes providing toString. 
* Refactor KindConnectionTests to split some tests, remove duplicates and fix some IO comparisons
* Refactored Resource to use an internal state machine like IO, in order to be able to do tail recursion over its bindings to avoid stack overflows.
* Simplify stack overflow test for MVar


arrow-fx/src/main/kotlin/arrow/fx/MVar.kt
```diff
@@ -168,7 +168,7 @@ interface MVar<F, A> {
   fun tryTake(): Kind<F, Option<A>>
 
   /**
-   * Tries reading the current value, or blocks until there is a value available.
+   * Reads the current value without touching it, assuming there is one, or otherwise it blocks until there is a value available.
```
nomisRev:
```suggestion
   * Reads the current value without emptying the MVar, assuming there is one, or otherwise it blocks until there is a value available.
```

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
```diff
@@ -72,7 +74,7 @@ class MVarTest : UnitSpec() {
           val bb = f2.join().bind()
 
           setOf(aa, bb)
-        }.equalUnderTheLaw(IO.just(setOf(10, 20)), EQ())
+        }.shouldBeEq(IO.just(setOf(10, 20)), EQ())
```
nomisRev: Implementation of `equalUnderTheLaw` should change to `shouldBeEq` instead. Much bigger improvement since that'll fix the error message for all currently using `equalUnderTheLaw`.

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
```diff
@@ -260,9 +263,9 @@ class MVarTest : UnitSpec() {
           val channel = !mvar.empty<Int>()
           val (count, reads, writes) = testStackSequential(channel)
           !writes.fork()
-          !reads
-          !effect { reads shouldBe count }
-        }.equalUnderTheLaw(IO.unit, EQ())
+          val r = !reads
+          !effect { r shouldBe count }
```
nomisRev: This seems like the only relevant fix?

arrow-fx-test/src/main/kotlin/arrow/test/laws/Law.kt
```diff
@@ -24,7 +24,7 @@ fun <A> A.shouldBeEq(b: A, eq: Eq<A>): Unit = this should matchUnderEq(eq, b)
 
 fun <A> matchUnderEq(eq: Eq<A>, b: A) = object : Matcher<A> {
   override fun test(value: A): Result {
-    return io.kotlintest.Result(eq.run { value.eqv(b) }, "Expected: $b but found: $value", "$b and $value should be equal")
+    return Result(value.equalUnderTheLaw(b, eq), "Expected: $b but found: $value", "$b and $value should be equal")
```
aballano: @nomisRev I actually did the opposite, make `shouldBeEq` above depend on `equalUnderTheLaw`

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
```diff
@@ -211,18 +213,19 @@ class MVarTest : UnitSpec() {
...
-        mvar.just(Option(0)).flatMap(::exec)
-          .unsafeRunSync() shouldBe count.toLong() * (count - 1) / 2
+        mvar.just(Option(0))
+          .flatMap(::exec)
+          .shouldBeEq(IO.just(count.toLong() * (count - 1) / 2), EQ())
       }
 
       "$label - producer-consumer parallel loop" {
-        fun producer(ch: Channel<Int>, list: List<Int>): IO<Unit> =
+        fun producer(ch: Channel<Long>, list: List<Long>): IO<Unit> =
```
nomisRev: Why the change to `Long`?

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
aballano: see https://github.com/arrow-kt/arrow-fx/pull/46#discussion_r383931087

arrow-fx-test/src/main/kotlin/arrow/test/laws/Law.kt
aballano: The problem is that the `forAll` tests expect a boolean at the end, which is provided by the `equalUnderTheLaw`, otherwise for a explicit check or throw the `shouldXXX` should be used due to convention.

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
aballano: Because the EQ generated was trying to compare an Int and a Long and therefore failing

arrow-fx-test/src/main/kotlin/arrow/test/laws/Law.kt
nomisRev: Not really, since that was already the case. `equalUnderTheLaw == eq.run { eqv(b) }` and you substituted it in `matchUnderEq` which means nothing changed. Subsitution law ;)

nomisRev:
```kotlin
fun <A> A.equalUnderTheLaw(b: A, eq: Eq<A>): Boolean =
  this should matchUnderEq(eq, b).let { true }

fun <A> A.shouldBeEq(b: A, eq: Eq<A>): Unit = this should matchUnderEq(eq, b)

fun <A> matchUnderEq(eq: Eq<A>, b: A) = object : Matcher<A> {
  override fun test(value: A): Result {
    return io.kotlintest.Result(eq.run { value.eqv(b) }, "Expected: $b but found: $value", "$b and $value should be equal")
```

nomisRev: This will make sure that all failing `forAll` fail with a proper message `"Expected: $b but found: $value", "$b and $value should be equal"`, instead of currently `"Expected: true but found: false`.

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
nomisRev: I see.

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
aballano: Yep, only this nit and the Int -> Long below :)

arrow-fx-test/src/main/kotlin/arrow/test/laws/Law.kt
```diff
@@ -18,13 +18,13 @@ fun throwableEq() = Eq { a: Throwable, b ->
 data class Law(val name: String, val test: suspend TestContext.() -> Unit)
 
 fun <A> A.equalUnderTheLaw(b: A, eq: Eq<A>): Boolean =
-  eq.run { eqv(b) }
+  shouldBeEq(b, eq).let { true }
```
aballano: Seems that this change unveiled some tests that weren't doing any verification with forAll or shouldBe and were failing 😓 cc @1Jajen1 @nomisRev 


arrow.fx.ResourceTest > Monad Laws: stack safe FAILED
arrow.fx.MVarTest > CancelableMVar - stack overflow test FAILED
arrow.fx.internal.TimeoutException: Duration(amount=5, timeUnit=SECONDS)
arrow.fx.MVarTest > CancelableMVar - producer-consumer parallel loop FAILED
arrow.fx.MVarTest > UncancelableMVar - concurrent take and put FAILED
arrow.fx.MVarTest > UncancelableMVar - stack overflow test FAILED
arrow.fx.MVarTest > UncancelableMVar - producer-consumer parallel loop FAILED
arrow.fx.IOTest > MonadDefer laws: stack safety over repeated left binds FAILED
arrow.fx.IOTest > MonadDefer laws: stack safety over repeated attempts FAILED
arrow.fx.SemaphoreTest > CancelableSemaphore - tryAcquire with available permits FAILED
arrow.fx.SemaphoreTest > UncancelableSemaphore - tryAcquire with available permits FAILED
arrow.fx.EqTest > Should fail for pure non-equal values FAILED

aballano:
Current test status:

- [ ] arrow.fx.ResourceTest > Monad Laws: stack safe FAILED
- [ ] arrow.fx.MVarTest > CancelableMVar - stack overflow test FAILED
- [ ] arrow.fx.internal.TimeoutException: Duration(amount=5, timeUnit=SECONDS)
- [ ] arrow.fx.MVarTest > CancelableMVar - producer-consumer parallel loop FAILED
- [ ] arrow.fx.MVarTest > UncancelableMVar - concurrent take and put FAILED
- [ ] arrow.fx.MVarTest > UncancelableMVar - stack overflow test FAILED
- [ ] arrow.fx.MVarTest > UncancelableMVar - producer-consumer parallel loop FAILED
- [ ] arrow.fx.IOTest > MonadDefer laws: stack safety over repeated left binds FAILED
- [ ] arrow.fx.IOTest > MonadDefer laws: stack safety over repeated attempts FAILED
- [ ] arrow.fx.SemaphoreTest > CancelableSemaphore - tryAcquire with available permits FAILED
- [ ] arrow.fx.SemaphoreTest > UncancelableSemaphore - tryAcquire with available permits FAILED
- [x] arrow.fx.EqTest > Should fail for pure non-equal values FAILED

arrow-fx/src/test/kotlin/arrow/fx/ResourceTest.kt
```diff
@@ -59,16 +55,18 @@ class ResourceTest : UnitSpec() {
...
   override fun <A> Kind<ResourcePartialOf<ForIO, Throwable>, A>.eqK(other: Kind<ResourcePartialOf<ForIO, Throwable>, A>, EQ: Eq<A>): Boolean =
     (this.fix() to other.fix()).let {
-      val ls = it.first.invoke { IO.just(1) }.fix()
-      val rs = it.second.invoke { IO.just(1) }.fix()
+      val ls = it.first.use { IO.just(it) }.fix().attempt()
+      val rs = it.second.use { IO.just(it) }.fix().attempt()
       val compare = IO.applicative().mapN(ls, rs) { (l, r) -> l == r }.fix()
 
-      compare.unsafeRunTimed(5.seconds) == Some(true)
+      compare.unsafeRunTimed(10.seconds) == Some(true)
```
nomisRev: Are the test running so much slower?

We shouldn't use `unsafeRunTimed`, it's not a safe alternative for timing out. It limits the time any `async` boundary can take, so infinite `flatMap`s never time-out. We should forward to `IO.eq` here instead.

arrow-fx/src/test/kotlin/arrow/fx/ResourceTest.kt
```diff
@@ -59,16 +55,18 @@ class ResourceTest : UnitSpec() {
...
 }
 
 private fun Resource.Companion.genK() = object : GenK<ResourcePartialOf<ForIO, Throwable>> {
-  override fun <A> genK(gen: Gen<A>): Gen<Kind<ResourcePartialOf<ForIO, Throwable>, A>> = gen.map {
-    Resource.just(it, IO.bracket())
-  }
+  override fun <A> genK(gen: Gen<A>): Gen<Kind<ResourcePartialOf<ForIO, Throwable>, A>> =
+    Gen.oneOf(
+      gen.map { just(it, IO.bracket()) },
+      IO.genK().genK(gen).map { it.liftF(IO.bracket()) }
```
nomisRev: We should invoke all cases of the `sealed class Resource` here, no?

arrow-fx/src/test/kotlin/arrow/fx/SemaphoreTest.kt
```diff
@@ -34,7 +34,7 @@ class SemaphoreTest : UnitSpec() {
       "$label - tryAcquire with available permits" {
         val n = 20
         semaphore(20).flatMap { s ->
-          (0 until n).toList().traverse(IO.applicative()) { s.acquire() }.flatMap {
+          (0 until n).toList().traverse(IO.applicative()) { IO.unit }.flatMap {
```
nomisRev: This test should probably be simplified. No reason to work with `20` permits and `traverse` just to test `tryAqcuire` with permits left. Ideally, we should use a `NonZeroPosInt` property test here instead and pass it to `semaphore(..)`.

arrow-fx/src/main/kotlin/arrow/fx/Resource.kt
```diff
@@ -148,7 +151,7 @@ inline fun <F, E, A> ResourceOf<F, E, A>.fix(): Resource<F, E, A> =
  * All three programs do exactly the same with varying levels of simplicity and overhead. `Resource` uses `Bracket` under the hood but provides a nicer monadic interface for creating and releasing resources in order, whereas bracket is great for one-off acquisitions but becomes more complex with nested resources.
  *
  **/
-interface Resource<F, E, A> : ResourceOf<F, E, A> {
+sealed class Resource<F, E, A> : ResourceOf<F, E, A> {
```
nomisRev: > Refactored Resource to use an internal state machine like IO, in order to be able to do tail recursion over its bindings to avoid stack overflows.

This refactor was needed to fix `tailRecM`?

arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
```diff
@@ -181,48 +183,32 @@ class MVarTest : UnitSpec() {
...
-            !f2.join()
-          }
-        }
-
-        mvar.just(Option(0)).flatMap(::exec)
-          .unsafeRunSync() shouldBe count.toLong() * (count - 1) / 2
+        mvar.just(0)
+          .flatMap { it.read() }
+          .flatMap { ioMvarLoop(it) }
+          .flatMap { it.read() }.unsafeRunSync() shouldBe size
```
nomisRev: This is not the equivalent. I checked locally and there seems to be a bug in `CancelableMVar`.

I refactored `UncancelableMVar` to fix bugs and refactored it to use `IQueue` added with the `Queue` implementation. Doing the same refactor for `UncancelableMVar` might unveil the bug there too.

arrow-fx/src/test/kotlin/arrow/fx/ResourceTest.kt
aballano: Not really, just forgot to remove this after testing :D

arrow-fx/src/test/kotlin/arrow/fx/ResourceTest.kt
aballano: There are 2 and the third one is `flatMap` which is already tested through, so I don't think we need it here, no?

arrow-fx/src/main/kotlin/arrow/fx/Resource.kt
aballano: This refactor allows to define use from fold and deal with every case individually, so I'd say yes

arrow-fx/src/test/kotlin/arrow/fx/SemaphoreTest.kt
aballano: This seems to fail, maybe we should just revisit them later on and improve

OutOfMemoryError: Java heap space

arrow-fx/src/main/kotlin/arrow/fx/Resource.kt
```diff
@@ -165,57 +168,109 @@ interface Resource<F, E, A> : ResourceOf<F, E, A> {
...
-   *   val program = Resource(::acquireResource, ::releaseResource, IO.bracket()).invoke {
+   *   val program = Resource(::acquireResource, ::releaseResource, IO.bracket()).use {
    *     IO { println("Expensive resource under use! $it") }
    *   }
    *   //sampleEnd
    *   program.fix().unsafeRunSync()
    * }
    * ```
    */
-  operator fun <C> invoke(use: (A) -> Kind<F, C>): Kind<F, C>
```
nomisRev: This function should be `@Deprecated` in `0.10.x` and removed in `0.11.0`.
`Resource` is already available in the latest stable version `0.10.4`

arrow-fx/src/main/kotlin/arrow/fx/internal/LinkedMap.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import arrow.core.Tuple2
+import arrow.core.getOption
+import arrow.core.toT
+
+/**
+ * A Map which tracks the insertion order of entries, so that entries may be
+ * traversed in the order they were inserted.
+ */
+internal class LinkedMap<K, V>(
```
nomisRev: Can we use [`LinkedHashMap`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-map/index.html#kotlin.collections.LinkedHashMap) from `kotlin.collections` instead?

arrow-fx/src/main/kotlin/arrow/fx/internal/UncancellableMVar.kt
```diff
@@ -64,7 +64,7 @@ internal class UncancellableMVar<F, A> private constructor(initial: State<A>, pr
         var first: Listener<A>? = null
         val update: State<A> =
           if (takes.isEmpty()) {
-            WaitForTake(a, IQueue.empty())
+            State(a)
```
nomisRev: I moved from `State(a)` to more explicit `WaitForTake` recently since I felt that `State` seriously hindered readability.

https://github.com/arrow-kt/arrow/pull/2023/files#diff-14d465adcba579f7d8df401c3f1b3af4L63

arrow-fx/src/main/kotlin/arrow/fx/internal/UncancellableMVar.kt
```diff
@@ -135,7 +135,7 @@ internal class UncancellableMVar<F, A> private constructor(initial: State<A>, pr
         if (queue.isEmpty()) {
           if (stateRef.compareAndSet(current, State.empty())) {
             // Signals completion of `take`
-            onTake(Either.Right(value))
+            onTake(Right(value))
```
nomisRev: `Either.Right` is the constructor, `Right` goes through an additional method for no reason here.

arrow-fx/src/test/kotlin/arrow/fx/ResourceTest.kt
nomisRev: Here you can see how the `Gen` is build for `Resource` in Cats effect.

https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/main/scala/cats/effect/laws/discipline/arbitrary.scala#L118

It build `Allocate` from `Gen<Kind<F, A>>` and `Gen<Kind<F, Unit>>`, it builds `FlatMap` from the allocate `Gen` and another for `Suspend`.

The idea here is that if your `Kind<F, A>` for example `IO<A>` is a strong generator that also generates `IO` at depths etc so it can cover all its cases.

arrow-fx/src/main/kotlin/arrow/fx/internal/Utils.kt
```diff
@@ -300,3 +300,7 @@ internal fun <A> asyncContinuation(ctx: CoroutineContext, cc: (Either<Throwable,
  */
 internal fun <A> IOForkedStart(fa: IOOf<A>, ctx: CoroutineContext): IO<A> =
   IO.Bind(IO.ContinueOn(IO.unit, ctx)) { fa.fix() }
+
+@Suppress("UNCHECKED_CAST")
+internal fun <A, B> A.unsafeCast(): B =
```
nomisRev: I'm personally not a fan of this, and I think we should be consistent in the Arrow Fx codebase when doing this.

Currently it's hard coded casting everywhere, which I prefer so you can more explicitly see the types throughout the code (which is not easy to read or follow).

cc\\ @raulraja @pakoito

arrow-fx/src/main/kotlin/arrow/fx/internal/LinkedMap.kt
aballano: Nope AFAIK :D Actually the current MVar has already that kind of list

arrow-fx/src/main/kotlin/arrow/fx/internal/UncancellableMVar.kt
aballano: Indeed, it actually goes through 2 additional functions :/

arrow-fx/src/main/kotlin/arrow/fx/Resource.kt
```diff
@@ -203,31 +206,31 @@ sealed class Resource<F, E, A> : ResourceOf<F, E, A> {
...
 
-  private fun <F, B> fold(
+  private fun <B> fold(
     onOutput: (A) -> Kind<F, B>,
     onRelease: (Kind<F, Unit>) -> Kind<F, Unit>
   ): Kind<F, B> {
     // Interpreter that knows how to evaluate a Resource data structure
     // Maintains its own stack for dealing with Bind chains
-    tailrec fun loop(current: Resource<F, E, Any>, stack: List<(Any) -> Resource<F, E, Any>>): Kind<F, Any?> =
+    tailrec fun loop(current: Resource<F, E, A>, stack: List<(A) -> Resource<F, E, A>>): Kind<F, B> =
```
nomisRev: 😍

PR #47, opened 2020-02-25
Change applicative instances to implement apEval and remove lazyAp

1Jajen1:
Changes from https://github.com/arrow-kt/arrow-core/pull/23 broke every arrow build with an applicative instance, here is the fix.

This non-deterministically failed MVar tests locally. Is this linked to #46 ?

### DIFF
--- a/arrow-fx-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
+++ b/arrow-fx-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
@@ -18,9 +18,6 @@ import arrow.core.value
 import arrow.fx.IO
 import arrow.fx.extensions.io.applicative.applicative
 import arrow.fx.extensions.io.concurrent.concurrent
-import arrow.mtl.typeclasses.ComposedApplicative
-import arrow.mtl.typeclasses.nest
-import arrow.mtl.typeclasses.unnest
 import arrow.test.generators.GenK
 import arrow.test.generators.functionAToB
 import arrow.test.generators.intSmall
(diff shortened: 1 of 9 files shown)


pakoito:
This...may need to be rebased. +55 −1,842

1Jajen1:
> This...may need to be rebased. +55 −1,842

This deletes a bunch of files, basically all laws which are already in `arrow-core-test` so they were duplicates. Should have probably done this in another pr though, but this way it confirms that the split approach works.

pakoito:
I'd say splitsies

1Jajen1:
> I'd say splitsies

I'll try tomorrow but I am not sure that'll work. Current master should be broken by the change in core. And I would not want to merge that without some sort of ci...

rachelcarmena:
Hi @1Jajen1 , thank you so much!! Please, could this PR just change `apEval` without the refactoring. I'm not sure about the idea of using `arrow-core-test` here and we could decide about it in another PR. Or is it necessary to make `apEval` change? Thanks again!!

1Jajen1:
> Hi @1Jajen1 , thank you so much!! Please, could this PR just change `apEval` without the refactoring? I'm not sure about the idea of using `arrow-core-test` here and we could decide about it in another PR. Or is it necessary to make `apEval` change? Thanks again!!

Part of it is. Some laws use mtl and they broke the build and needed to change. Because it was little to no effort removing the rest of it as well I did that in this pr as well. I'll see how much of this is needed for the build to work and then do the rest in another pr

arrow-fx-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
```diff
@@ -46,6 +43,17 @@ class TIF {
...
+
+fun <F, G, A> Kind<F, Kind<G, A>>.nest(): Kind<Nested<F, G>, A> = this as Kind<Nested<F, G>, A>
+fun <F, G, A> Kind<Nested<F, G>, A>.unnest(): Kind<F, Kind<G, A>> = this as Kind<F, Kind<G, A>>
+
+fun <F, G> ComposedApplicative(apF: Applicative<F>, apG: Applicative<G>): Applicative<Nested<F, G>> = object : Applicative<Nested<F, G>> {
+  override fun <A, B> Kind<Nested<F, G>, A>.ap(ff: Kind<Nested<F, G>, (A) -> B>): Kind<Nested<F, G>, B> =
+    apF.run { unnest().ap(ff.unnest().map { gf -> { ga: Kind<G, A> -> apG.run { ga.ap(gf) } } }).nest() }
+  override fun <A> just(a: A): Kind<Nested<F, G>, A> = apF.just(apG.just(a)).nest()
+}
+
```
1Jajen1: This will be removed in a cleanup pass on arrow-fx-test anyway, for now this is duplicate code...

aballano:
Let's merge and I'll take a look at the tests on MVar again

PR #51, opened 2020-02-26
Clean up arrow-fx-test

1Jajen1:
Removes a bunch of duplicate code.

This fails building with gradle locally because it can't find the generators for some reason. Intellij fails to shows this as an error and I don't even understand why it'd fail^^

Edit: I also changed `arrow-fx-test` to be published the same way as `arrow-core-test`


arrow-fx-test/src/main/kotlin/arrow/test/util/LawUtils.kt
```diff
@@ -7,19 +7,12 @@ import arrow.test.generators.tuple5
...
 fun throwableEq() = Eq { a: Throwable, b ->
   a::class == b::class && a.message == b.message
 }
-
-data class Law(val name: String, val test: suspend TestContext.() -> Unit)
-
-fun <A> A.equalUnderTheLaw(b: A, eq: Eq<A>): Boolean =
-  eq.run { eqv(b) }
-
 fun <A> A.shouldBeEq(b: A, eq: Eq<A>): Unit = this should matchUnderEq(eq, b)
```
nomisRev: Shouldn't all this file belong in arrow-core-test?

arrow-fx/src/test/kotlin/arrow/fx/FiberTest.kt
```diff
@@ -24,14 +25,14 @@ class FiberTest : UnitSpec() {
 
   init {
     fun EQ(): Eq<FiberOf<ForIO, Int>> = object : Eq<FiberOf<ForIO, Int>> {
```
nomisRev: Can we move these to `arrow-fx-test` as well and unfix them to `ForIO`?

1Jajen1:
> Looking great so far! Do you mind if I give it a shot to finish this PR @1Jajen1?

Fine by me. This fails to compile for reasons I don't understand^^ Wanted to take a look at it this weekend, but if you want to try it go ahead :)

arrow-fx-test/src/main/kotlin/arrow/test/util/LawUtils.kt
1Jajen1: Yes, had this here because I wanted to see if this even compiles before moving it

1Jajen1:
This is continued in #90

PR #59, opened 2020-02-27
Use GenK from arrow-core-test

aballano:
Same as in https://github.com/arrow-kt/arrow-ui/pull/15


aballano:
only the current flaky test failed, all others succeeded, so I'd say we can merge :D

> arrow.fx.MVarTest > UncancelableMVar - concurrent take and put FAILED

PR #67, opened 2020-02-27
add a SemigroupK IO instance

abendt:
this continues https://github.com/arrow-kt/arrow/pull/2043
after the split of the arrow monorepo


1Jajen1:
Thanks for adding this^^ Combining `IO` with this might be a niche use case, but good to have anyway.

Btw: Haskell `IO` has an `Alternative` instance (which is not lawful unless you consider all exceptions equal). We could add it as well, but that would mean that any `raiseException(x) == raiseException(y)` even if `x != y`. Thoughts? It would mean that `IO` gets access to `Alternative` methods like `altSum` and others which may be helpful.

PR #73, opened 2020-03-02
Conf: control transitive dependencies

rachelcarmena:


arrow-fx-test/build.gradle
```diff
@@ -11,15 +11,15 @@ apply from: "$SUBPROJECT_CONF"
...
     compile project(":arrow-fx")
-    compile "io.arrow-kt:arrow-core-test:$VERSION_NAME"
-    compile "io.arrow-kt:arrow-mtl-data:$VERSION_NAME"
-    compile "io.arrow-kt:arrow-free-data:$VERSION_NAME"
-    compile "io.arrow-kt:arrow-recursion-data:$VERSION_NAME"
+    compile "io.arrow-kt:arrow-core-test:$VERSION_NAME", excludeArrow
+    compile "io.arrow-kt:arrow-mtl:$VERSION_NAME", excludeArrow
+    compile "io.arrow-kt:arrow-mtl-data:$VERSION_NAME", excludeArrow
+    compile "io.arrow-kt:arrow-free:$VERSION_NAME", excludeArrow
+    compile "io.arrow-kt:arrow-free-data:$VERSION_NAME", excludeArrow
```
aballano: since you're adding more dependencies here, i assume those are the implicitly passed by the others above, right? Would it mean that we should be able to quickly switch to `implementation` in gradle files now? 😍

rachelcarmena: Thanks for the review @aballano !! I improved some configurations with the new Gradle methods, however it's pending for Arrow Fx (a lot of time to run tests). I'll improve it next weeks. Thanks!!

rachelcarmena:
Thanks @aballano !! :raised_hands:

Issue #31, opened 2020-02-20
KTLint: enable again for all the modules


rachelcarmena:
With the new repository organization, KTLint has been disabled for some modules and:

```
ktlintTestSourceSetCheck.enabled = false
```

has been added in these files:
 
*    `arrow-fx-kotlinx-coroutines/build.gradle`
*    `arrow-fx/build.gradle`

It would be necessary to remove it and fix the issues.

rachelcarmena:
Related to https://github.com/arrow-kt/arrow-ank/issues/9

PR #79, opened 2020-03-04
Enable ktlint

danimontoya:
This PR enables ktLint #31 .

### DIFF
--- a/arrow-fx/src/test/kotlin/arrow/fx/RefTest.kt
+++ b/arrow-fx/src/test/kotlin/arrow/fx/RefTest.kt
@@ -18,7 +18,7 @@ import io.kotlintest.shouldBe
 class RefTest : UnitSpec() {
 
   init {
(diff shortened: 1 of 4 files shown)


arrow-fx/src/test/kotlin/arrow/fx/RefTest.kt
```diff
@@ -18,7 +18,7 @@ import io.kotlintest.shouldBe
 class RefTest : UnitSpec() {
 
   init {
-    fun <F> MonadDefer<F>.tests(EQF: EqK<F>, RF: RefFactory<F>): Unit {
```
rachelcarmena: Congrats @danimontoya !! You got it!! Now we could undo these changes. They aren't API but an example for other functions :+1:

danimontoya: Sure!

danimontoya: Thanks :)

PR #80, opened 2020-03-04
Refactor parMapN versions

### LINKED ISSUE #78: ["Request"] Remove CoroutineContext from Concurrent's parMapN and align API with Applicative
* The CoroutineContext is very unfriendly for discovering the API since we model it as IO.parMapN

Since it touches the same file, it would be a good opportunity to unify the API with Applicative's mapN in the following:
* Add parTupledN overloads for parMapN which use ::TupleN as the map function
* Fix parMapN to use Tuple like mapN to get rid of the `{ _, _, _, _, _ -> }` effect

_As suggested by @nomisRev [in the Koltin slack](https://kotlinlang.slack.com/archives/C8UK6RTHU/p1583252379033300?thread_ts=1583251480.030700&cid=C8UK6RTHU)_

aballano:
Fixes #78 

Deprecate parMapN ext fun, add explicit param
Add parTupledN versions

### DIFF
--- a/arrow-fx-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
+++ b/arrow-fx-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
@@ -5,6 +5,7 @@ import arrow.core.Left
 import arrow.core.ListK
 import arrow.core.Right
 import arrow.core.Tuple2
+import arrow.core.Tuple6
 import arrow.core.extensions.eq
 import arrow.core.extensions.listk.traverse.traverse
 import arrow.core.extensions.tuple2.eq.eq
@@ -624,7 +625,7 @@ object ConcurrentLaws {
       }.equalUnderTheLaw(just(a + b + c), EQ)
     }
 
-  fun <F> Concurrent<F>.parMapStartsAllAtSameTime(EQ: Eq<Kind<F, List<Long>>>) {
+  fun <F> Concurrent<F>.parMapStartsAllAtSameTime(EQ: Eq<Kind<F, Any>>) {
     val order = mutableListOf<Long>()
 
     fun makePar(num: Long) = sleep((num * 100).milliseconds).map {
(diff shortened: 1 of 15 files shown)


arrow-fx/src/main/kotlin/arrow/fx/IOParMap.kt
```diff
@@ -7,60 +7,259 @@ import arrow.core.Option
...
 import arrow.fx.internal.Platform
 import kotlin.coroutines.CoroutineContext
-import arrow.core.extensions.option.applicativeError.handleError
-import arrow.core.internal.AtomicBooleanW
 
-/** Mix-in to enable `parMapN` 2-arity on IO's companion directly. */
+/** Mix-in to enable `parMapN` N-arity on IO's companion directly. */
 interface IOParMap {
 
+  @Deprecated("This API is not consistent with others within Arrow, see version with Tuple instead of function params")
```
danimontoya: No need for `ReplaceWith` here?

aballano: Unfortunately it cannot be autocorrected AFAIK, but the signature is almost identical

arrow-fx-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -624,18 +625,18 @@ object ConcurrentLaws {
       }.equalUnderTheLaw(just(a + b + c), EQ)
     }
 
-  fun <F> Concurrent<F>.parMapStartsAllAtSameTime(EQ: Eq<Kind<F, List<Long>>>) {
-    val order = mutableListOf<Long>()
+  fun <F> Concurrent<F>.parMapStartsAllAtSameTime(EQ: Eq<Kind<F, Any>>) {
```
nomisRev: 😱 `Eq<Kind<F, Any>>` should be `Eq<Kind<F, Tuple6<Int, Int, Int, Int, Int, Int>>>`

PR #82, opened 2020-03-04
Update Arrow "cancel" wording for consistency 

aballano:
Brings https://github.com/arrow-kt/arrow/pull/2050 to 0.10.5 with deprecation over functions and types


rachelcarmena:
Hi! Sorry for not reviewing this PR before... this change is already done for `0.11.0-SNAPSHOT` by @nomisRev . I don't know if we want to wait for that version to do this change...

rachelcarmena:
**Arrow Incubator** and **Arrow Optics** were also changed for that version because this improvement impacts to them. 

BTW, the issue wasn't created because of having double quotes in the PR title (it's already fixed). It will be created in a few minutes.

aballano:
> Hi! Sorry for not reviewing this PR before... this change is already done for `0.11.0-SNAPSHOT` by @nomisRev . I don't know if we want to wait for that version to do this change...

Not exactly, I merged the change into the release branch due to a mistake, the process should have been first this PR with the deprecations and then remove the deprecations in 0.11

> Arrow Incubator and Arrow Optics were also changed for that version because this improvement impacts to them.

We will take a look there as well, thank you!!

PR #90, opened 2020-03-06
Cleanup fx tests

danimontoya:
Continues the work done in #51 :
Removes a bunch of duplicate code.

This fails building with gradle locally because it can't find the generators for some reason. Intellij fails to shows this as an error and I don't even understand why it'd fail^^

Edit: I also changed arrow-fx-test to be published the same way as arrow-core-test


1Jajen1:
Is this a redo of #51 ? If so I'll close that and stop trying to figure out what was broken since this one actually compiles^^ :+1:

danimontoya:
Yes sorry I did not tell you yet.. but was being hard to make it work and updating #51 😅 
I'll ping you when its done to have a look! thanks!

1Jajen1:
> Yes sorry I did not tell you yet.. but was being hard to make it work and updating #51 sweat_smile
I'll ping you when its done to have a look! thanks!

No problem, just got a bit confused. Glad to see this compiling^^

arrow-fx-test/src/main/kotlin/arrow/test/eq/EqK.kt
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+fun IO.Companion.eqK() = object : EqK<ForIO> {
+  override fun <A> Kind<ForIO, A>.eqK(other: Kind<ForIO, A>, EQ: Eq<A>): Boolean = eq(EQ).run {
+    fix().eqv(other.fix())
+  }
+}
+
+fun <F, A> EQ(EQ: Eq<Kind<F, A>>): Eq<FiberOf<F, A>> = object : Eq<FiberOf<F, A>> {
```
nomisRev:
```suggestion
fun <F, A> Fiber.Companion.eq(EQ: Eq<Kind<F, A>>): Eq<FiberOf<F, A>> = object : Eq<FiberOf<F, A>> {
```

We can also add an `EqK` instance for `Fiber`.

arrow-fx-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -85,50 +78,28 @@ fun <E, A> Gen.Companion.either(genE: Gen<E>, genA: Gen<A>): Gen<Either<E, A>> {
...
+
+fun <A> Gen<A>.genEval(): Gen<Eval<A>> =
+  map { Eval.just(it) }
+
+fun IO.Companion.genK() = object : GenK<ForIO> {
+  override fun <A> genK(gen: Gen<A>): Gen<Kind<ForIO, A>> = Gen.oneOf(
+    gen.map(IO.Companion::just),
+    Gen.throwable().map(IO.Companion::raiseError)
+  )
+}
```
nomisRev: `GenK` of `Fiber` can also be moved here

arrow-fx-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -85,50 +78,28 @@ fun <E, A> Gen.Companion.either(genE: Gen<E>, genA: Gen<A>): Gen<Either<E, A>> {
...
-    override fun random(): Sequence<Ior<A, B>> =
-      (Gen.option(genA).random() to Gen.option(genB).random()).let { (ls, rs) ->
-        ls.zip(rs).filterMap {
-          Ior.fromOptions(it.first, it.second)
-        }
-      }
-  }
+fun <A, B> Gen.Companion.functionABToB(gen: Gen<B>): Gen<(A, B) -> B> = gen.map { b: B -> { _: A, _: B -> b } }
+
+fun <A> Gen<A>.genEval(): Gen<Eval<A>> =
```
nomisRev: `Eval` comes from `arrow.core`.

arrow-fx-test/src/main/kotlin/arrow/test/laws/Law.kt
```diff
@@ -7,17 +7,15 @@ import arrow.test.generators.tuple5
 import arrow.typeclasses.Eq
 import io.kotlintest.Matcher
 import io.kotlintest.Result
-import io.kotlintest.TestContext
 import io.kotlintest.properties.Gen
 import io.kotlintest.should
 import io.kotlintest.shouldNot
 
+// TODO: check to move to arrow-core-test
 fun throwableEq() = Eq { a: Throwable, b ->
```
nomisRev: I am okay leaving this in `arrow-fx-test` since effects imply `Throwable`, and comparing `Throwable` can be dependent on use-case. Let's move this with the other `Eq` instances tho.

I think this is a fair implementation to leave public in `arrow-fx-test`, but we should add a clear KDoc that comparing `Throwable` is not safe due to their structure (stacktrace), so we structurally compare type and message instead.

arrow-fx/src/test/kotlin/arrow/fx/FiberTest.kt
```diff
@@ -23,15 +24,10 @@ import io.kotlintest.properties.Gen
 class FiberTest : UnitSpec() {
 
   init {
-    fun EQ(): Eq<FiberOf<ForIO, Int>> = object : Eq<FiberOf<ForIO, Int>> {
-      override fun FiberOf<ForIO, Int>.eqv(b: FiberOf<ForIO, Int>): Boolean = EQ<Int>().run {
-        fix().join().eqv(b.fix().join())
-      }
-    }
 
     fun EQK() = object : EqK<FiberPartialOf<ForIO>> {
```
nomisRev: See other comment, move to `arrow-fx-test`.

arrow-fx-test/build.gradle
```diff
@@ -8,15 +9,13 @@ plugins {
...
 apply from: "$SUBPROJECT_CONF"
+apply from: "$PUBLISH_CONF"
 
 dependencies {
     compile project(":arrow-fx")
-    compile "io.arrow-kt:arrow-core-test:$VERSION_NAME", excludeArrow
+    api "io.arrow-kt:arrow-core-test:$VERSION_NAME", excludeArrow
     compile "io.arrow-kt:arrow-mtl:$VERSION_NAME", excludeArrow
-    compile "io.arrow-kt:arrow-mtl-data:$VERSION_NAME", excludeArrow
     compile "io.arrow-kt:arrow-free:$VERSION_NAME", excludeArrow
```
nomisRev:
```suggestion
```

I don't think we use free in `arrow-fx-test`

arrow-fx-test/build.gradle
```diff
@@ -8,15 +9,13 @@ plugins {
 }
 
 apply from: "$SUBPROJECT_CONF"
+apply from: "$PUBLISH_CONF"
 
 dependencies {
     compile project(":arrow-fx")
-    compile "io.arrow-kt:arrow-core-test:$VERSION_NAME", excludeArrow
+    api "io.arrow-kt:arrow-core-test:$VERSION_NAME", excludeArrow
     compile "io.arrow-kt:arrow-mtl:$VERSION_NAME", excludeArrow
```
nomisRev: We don't need mtl in `arrow-fx-test`

arrow-fx-test/build.gradle
```diff
@@ -8,15 +9,13 @@ plugins {
...
-    compile "io.arrow-kt:arrow-mtl-data:$VERSION_NAME", excludeArrow
     compile "io.arrow-kt:arrow-free:$VERSION_NAME", excludeArrow
-    compile "io.arrow-kt:arrow-free-data:$VERSION_NAME", excludeArrow
-    compile "io.arrow-kt:arrow-recursion-data:$VERSION_NAME", excludeArrow
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
     compile "org.jetbrains.kotlinx:kotlinx-coroutines-core:$KOTLINX_COROUTINES_VERSION"
```
nomisRev: Created #93 for the removal of this dependency. This is low priority so not blocking for 0.10.x

arrow-fx-test/src/main/kotlin/arrow/test/eq/EqK.kt
```diff
@@ -0,0 +1,53 @@
+package arrow.test.eq
```
rachelcarmena: Hi! Please, if `arrow-fx-test` is going to be publish in OSS repository, replace `arrow.test` package by `arrow.fx.test` in order to avoid misunderstandings with other test artifacts. We'll do the same with `arrow-core-test` that is already published.

rachelcarmena: Thanks!! :raised_hands:

nomisRev:
Unblocks https://github.com/arrow-kt/arrow-incubator/pull/63

PR #104, opened 2020-03-12
Check binary compatibility before releasing 0.10.5

danimontoya:
Check binary compatibility


arrow-fx/build.gradle
```diff
@@ -9,6 +9,10 @@ plugins {
     id "ru.vyarus.animalsniffer"
 }
 
+apiValidation {
```
rachelcarmena: Good job @danimontoya :clap: Just a comment. Could `apiValidation` section be located in `build.gradle`?

danimontoya: I guess you mean in the root `build.gradle`? I've been playing with it and no, I couldn't =\
so I had to add it to arrow-fx and worked as I was expecting if I would have added it to the root build.gradle

danimontoya: you right @rachelcarmena ! just moved it to the root one! thanks!

rachelcarmena:
> LGTM although we should create a ticket to add the api check change to the CI

Or maybe it can be included in this PR. Just add a new step in `.github/workflows/build_arrow-fx.yml` at the end:

```
    - name: Check the API
      run: ./gradlew apiCheck
```

danimontoya:
Oh thats awesome! But adding it now can create some issues.. its better to create a separate issue and tackle it together with adding the possible ignore projects and ignore packages!!

rachelcarmena:
> Oh thats awesome! But adding it now can create some issues.. its better to create a separate issue and tackle it together with adding the possible ignore projects and ignore packages!!

Good idea!!! :raised_hands:

danimontoya:
Task will be continued here: https://github.com/arrow-kt/arrow-fx/issues/106

PR #132, opened 2020-03-26
Fix doc generation and validation failures

rachelcarmena:
PR on `release/0.11.0` branch to fix failures when generating and validating the documentation from #130.


arrow-fx/src/main/kotlin/arrow/fx/Promise.kt
```diff
@@ -103,7 +103,7 @@ interface Promise<F, A> {
    *       p.complete(2)
    *     }
    *   }.attempt().unsafeRunSync() ==
-   *     IO.raiseError<Int>(Promise.AlreadyFulfilled).attempt().unsafeRunSync()
+   *     IO.raiseException<Int>(Promise.AlreadyFulfilled).attempt().unsafeRunSync()
```
pakoito: This...doesn't seem right. The generic should be of the type of `Promise.AlreadyFulfilled` instead, right?

pakoito: Oh nevermind, it's an attempt. Ye.

nomisRev:
@rachelcarmena I can continue this. I was also looking at it during the weekend:D

rachelcarmena:
> @rachelcarmena I can continue this. I was also looking at it during the weekend:D

Thanks @nomisRev !!! :raised_hands: Just 2 remaining failures :clap: Thanks for your help!!

settings.gradle
```diff
@@ -1,7 +1,6 @@
 rootProject.name = 'arrow-fx-lib'
 
 include 'arrow-fx'
-include 'arrow-docs'
```
nomisRev: Whoops

rachelcarmena: Don't worry @nomisRev :heart:

PR #134, opened 2020-03-31
New BIO APIs for unhappy paths

pakoito:
After consulting with the community about less scary and more familiar API names, we have landed in two terms acquired from the young people's lingo.

Instead of throwing we will `yeet` to signify that the `Exceptions` will be reaching far up in the callstack.

We believe `kobe` perfectly represents accurate and elegant error signaling aming directly to your domain.

I hope adopting these terms will drive us away from the more academic lexicon the rest of the library uses.


pakoito:
This was a good April's Fools. Closing now.

Stay safe y'all!

PR #135, opened 2020-03-31
Fix package for eq

rachelcarmena:
PR on `release/0.11.0` branch.

cc @nomisRev , we left `eq.kt` outside `arrow.fx.test` package.


rachelcarmena:
Thanks for the review @nomisRev !!

PR #148, opened 2020-04-23
Reactor concurrent instances

nomisRev:
Adds `Mono` & `Flux` instances for the `Concurrent` typeclass.


arrow-fx-reactor/src/test/kotlin/arrow/fx/FluxKTest.kt
```diff
@@ -164,26 +173,22 @@ class FluxKTest : UnitSpec() {
...
-        throw IllegalStateException()
-      } catch (err: Throwable) {
-        err
-      }
-
-      errA == errB
-    }
+  override fun FluxKOf<T>.eqv(b: FluxKOf<T>): Boolean {
+    val res1 = Try { value().timeout(Duration.ofSeconds(5)).blockFirst() }
+    val res2 = Try { b.value().timeout(Duration.ofSeconds(5)).blockFirst() }
```
aballano, line 178: I would avoid using Try as it'll only add more "weight" to removing it later

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
```diff
@@ -225,31 +264,29 @@ data class MonoK<out A>(val mono: Mono<out A>) : MonoKOf<A> {
...
-        val cb = { either: Either<Throwable, A> ->
-          either.fold(sink::error, sink::success)
-        }
-
-        val token = try {
-          fa(cb)
-        } catch (t: Throwable) {
-          cb(Left(t.nonFatalOrThrow()))
-          just(Unit)
+      Flux.create<A> { sink ->
```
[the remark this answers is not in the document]
nomisRev: Hehe. I was unable to implement this operator using `Mono` :/

It works fine through `Flux`. I'd opt to create a separate ticket for this, since it's an optimization rather than incorrect behavior.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/ForwardDisposable.kt
```diff
@@ -0,0 +1,70 @@
...
+ * This is useful for when you want to link the cancel lifecycle of an `Emitter`
+ * to running an [Disposable] stream.
+ *
+ * This is used in `bracket`, `guaranteeCase` & `asyncF`.
+ */
+internal class ForwardDisposable {
+
+  private val state = atomic(init)
+
+  fun cancel(): Disposable {
```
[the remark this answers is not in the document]
nomisRev: It's actually a bit trickier than that 😅
This doesn't really cancel anything. `ForwardDisposable` exposes 2 functions.

- `cancel`, which returns a `Disposable` that can be linked to an `Sink` using `onCancel`.
- `complete`, to atomically set the `Disposable` of the stream that is running within the coupled `Sink`.

I kept this the same as for `IO` since it's only used internally and not exposed.

arrow-fx-reactor/src/test/kotlin/arrow/fx/FluxKTest.kt
nomisRev: This is not correct anyway.. I'm going to rewrite this `Eq` instance to take into account all values within the `Flux`.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/common.kt
```diff
@@ -0,0 +1,7 @@
+package arrow.fx.reactor.extensions
+
+import arrow.core.Tuple2
+import java.util.function.BiFunction
+
+fun <A, B> tupled(): BiFunction<A, B, Tuple2<A, B>> =
```
pakoito, line 6: Scope it a bit so it doesn't appear on the global namespace. Internal should suffice IMO.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/dispatchers.kt
```diff
@@ -0,0 +1,36 @@
...
+internal val ComputationScheduler: CoroutineContext =
+  Schedulers.parallel().asCoroutineContext()
+
+internal val IOScheduler: CoroutineContext =
+  Schedulers.elastic().asCoroutineContext()
+
+fun Scheduler.asCoroutineContext(): CoroutineContext =
+  SchedulerContext(this)
+
+private class SchedulerContext(val scheduler: Scheduler) : AbstractCoroutineContextElement(ContinuationInterceptor), ContinuationInterceptor {
```
pakoito, line 19: I thought we had this already someplace else.

arrow-fx-reactor/src/test/kotlin/arrow/fx/FluxKTest.kt
```diff
@@ -164,26 +173,22 @@ class FluxKTest : UnitSpec() {
...
-
-      val errB = try {
-        b.value().blockFirst()
-        throw IllegalStateException()
-      } catch (err: Throwable) {
-        err
-      }
-
-      errA == errB
-    }
```
pakoito, line 186: Yikes

arrow-fx-reactor/src/test/kotlin/arrow/fx/FluxKTest.kt
```diff
@@ -164,26 +173,22 @@ class FluxKTest : UnitSpec() {
...
+        if (t1::class.java == java.util.concurrent.TimeoutException::class.java) throw t1
+        if (t2::class.java == java.util.concurrent.TimeoutException::class.java) throw t2
+        (t1::class.java == t2::class.java)
+      }, { false })
+    }, { v1 ->
+      res2.fold({ false }, {
+        v1 == it
+      })
+    })
+  }
```
pakoito, line 191: Yikes but a bit less

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/ForwardDisposable.kt
```diff
@@ -0,0 +1,70 @@
...
+      override fun dispose() =
+        loop(this)
+    }
+  }
+
+  fun complete(value: Disposable): Unit = state.value.let { current ->
+    when (current) {
+      is State.Active -> {
+        value.dispose()
+        throw IllegalStateException(current.toString())
```
pakoito, line 40: Is this the semantics in `reactor`?

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/ForwardDisposable.kt
```diff
@@ -0,0 +1,70 @@
...
+      }
+      is State.Empty -> if (current == init) {
```
pakoito, line 42: mmmmmmm `Disposable` does't have an identity. I'd use `===` to do reference comparison instead.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/ForwardDisposable.kt
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * [ForwardDisposable] allows you to link 2 [Disposable]s.
+ *
+ * This is useful for when you want to link the cancel lifecycle of an `Emitter`
+ * to running an [Disposable] stream.
+ *
+ * This is used in `bracket`, `guaranteeCase` & `asyncF`.
+ */
+internal class ForwardDisposable {
```
pakoito, line 15: Nit: interface and create instances in the companion object

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
```diff
@@ -97,33 +102,64 @@ data class MonoK<out A>(val mono: Mono<out A>) : MonoKOf<A> {
...
+            .subscribe({ emitter.error(e) }, { e2 -> emitter.error(Platform.composeErrors(e, e2)) })
+        }, {
+          Mono.defer { f(ExitCase.Completed).value() }
+            .subscribe({ emitter.success() }, { e -> emitter.error(e) })
+        }))
+    }.k()
+
+  fun fork(coroutineContext: CoroutineContext): MonoK<Fiber<ForMonoK, A>> =
+    coroutineContext.asScheduler().let { scheduler ->
+      Mono.create<Fiber<ForMonoK, A>> { emitter ->
```
pakoito, line 153: If the emitter is only used to return a value, it's better to use the blocking constructor. In `Rx` is something like `fromCompletable`.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
```diff
@@ -97,33 +102,64 @@ data class MonoK<out A>(val mono: Mono<out A>) : MonoKOf<A> {
...
+            Mono.defer { f(ExitCase.Completed).value() }
+              .subscribe({ emitter.success(a) }, { e -> emitter.error(e) })
+          }, { e ->
+          Mono.defer { f(ExitCase.Error(e)).value() }
+            .subscribe({ emitter.error(e) }, { e2 -> emitter.error(Platform.composeErrors(e, e2)) })
+        }, {
+          Mono.defer { f(ExitCase.Completed).value() }
+            .subscribe({ emitter.success() }, { e -> emitter.error(e) })
+        }))
+    }.k()
```
pakoito, line 149: 1. magic
2. pray
3. ???
4. cancellation

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
```diff
@@ -97,33 +102,64 @@ data class MonoK<out A>(val mono: Mono<out A>) : MonoKOf<A> {
...
-            })
-          } else {
-            throw e
-          }
-        }
-      } else sink.success(null)
-    })
+        },
+        emitter::error,
+        { }
```
pakoito, line 127: completion? do they need to be forwarded?

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
```diff
@@ -91,32 +93,45 @@ data class FluxK<out A>(val flux: Flux<out A>) : FluxKOf<A> {
...
-            })
-          } else {
-            throw e
-          }
-        }
-      }, sink::error, sink::complete)
-    })
+        },
+        { e: Throwable -> emitter.error(e) },
+        {}
```
pakoito, line 115: Same as mono, does completion need to be forwarded?

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
```diff
@@ -140,6 +155,18 @@ data class FluxK<out A>(val flux: Flux<out A>) : FluxKOf<A> {
       GA.run { f(a).map2Eval(eval) { Flux.concat(Flux.just<B>(it.a), it.b.flux).k() } }
     }.value()
 
+  fun fork(coroutineContext: CoroutineContext): FluxK<Fiber<ForFluxK, A>> =
+    coroutineContext.asScheduler().let { scheduler ->
+      Flux.create<Fiber<ForFluxK, A>> { emitter ->
+        val s: ReplayProcessor<A> = ReplayProcessor.create<A>()
+        val conn: reactor.core.Disposable = value()
+          .subscribeOn(scheduler)
```
pakoito, line 163: Nit: The scheduler is only use here, the `let` may be too extra

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
```diff
@@ -97,33 +102,64 @@ data class MonoK<out A>(val mono: Mono<out A>) : MonoKOf<A> {
...
+          Mono.defer { f(ExitCase.Error(e)).value() }
+            .subscribe({ emitter.error(e) }, { e2 -> emitter.error(Platform.composeErrors(e, e2)) })
+        }, {
+          Mono.defer { f(ExitCase.Completed).value() }
+            .subscribe({ emitter.success() }, { e -> emitter.error(e) })
+        }))
+    }.k()
+
+  fun fork(coroutineContext: CoroutineContext): MonoK<Fiber<ForMonoK, A>> =
+    coroutineContext.asScheduler().let { scheduler ->
```
pakoito, line 152: Nit: as in FluxK, this scheduler is only used in one place.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/dispatchers.kt
nomisRev: For RxJava.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/ForwardDisposable.kt
nomisRev: Actually `EMPTY_DISPOSABLE` exists as an `empty` reference iirc.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/ForwardDisposable.kt
nomisRev: Okay

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
nomisRev: Right!

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
nomisRev: No, the inner subcribed `Mono` calls the terminal operators.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
nomisRev: The inner one needs to call `completion` after this one completes...

Needs fixing.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
nomisRev: Run this `Mono`, and guarantee calling `Cancelled`, `Completed` or `Error`.
Wire `Disposable` running this `Mono` to `Mono.create` using `ForwardDisposable`.

No magic or praying 😛

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/ForwardDisposable.kt
nomisRev: Crutch

nomisRev:
@pakoito @aballano I removed `ForwardDisposable` since I was able to replace it with `Disposable.Composite` which also takes care of synchronization.

I addressed all comments besides the ones about `Eq`.
`Eq<ForFluxK>` is still incorrect atm, it should compare all values within the `Flux`.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
```diff
@@ -246,9 +248,9 @@ data class FluxK<out A>(val flux: Flux<out A>) : FluxKOf<A> {
 
     fun <A> asyncF(fa: ((Either<Throwable, A>) -> Unit) -> FluxKOf<Unit>): FluxK<A> =
       Flux.create { sink: FluxSink<A> ->
-        val cancellable = ForwardDisposable()
-        sink.onCancel(cancellable.cancel())
-        cancellable.complete(
+        val cancellable = Disposables.composite()
```
pakoito, line 251: Is this the one from reactor?

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/fluxk.kt
```diff
@@ -196,79 +197,107 @@ interface FluxKConcurrent : Concurrent<ForFluxK>, FluxKAsync {
...
       Flux.create<RacePair<ForFluxK, A, B>> { emitter ->
         val sa = ReplayProcessor.create<A>()
         val sb = ReplayProcessor.create<B>()
 
+        emitter.onCancel(forward)
+
         val dda = fa.value()
           .subscribeOn(scheduler)
           .subscribe(sa::onNext, sa::onError)
+          .also { forward.add(it) }
```
pakoito, line 212: NIT: `also(forward::add)`

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/fluxk.kt
```diff
@@ -196,79 +197,107 @@ interface FluxKConcurrent : Concurrent<ForFluxK>, FluxKAsync {
...
+    active: AtomicBooleanW,
     other2: ReactorDisposable,
     sink: FluxSink<RacePair<ForFluxK, A, B>>,
     err: Throwable
-  ) {
+  ): Unit = if (active.getAndSet(false)) {
     other2.dispose()
     sink.error(err)
-  }
+  } else Unit
```
pakoito, line 248: huh

I mean it's the most obvious one yet I'd never used it. I'd gone for some more clever yet crappier.

arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1211,13 +1211,13 @@ fun <E, A> IOOf<E, A>.onException(finalizer: (Throwable) -> IOOf<E, Unit>): IO<E
  * @param ctx [CoroutineContext] to execute the source [IO] on.
  * @return [IO] with suspended execution of source [IO] on context [ctx].
  */
-fun <E, A> IOOf<E, A>.fork(ctx: CoroutineContext = IO.dispatchers<Nothing>().default()): IO<E, Fiber<IOPartialOf<E>, A>> =
-  IO.async { cb ->
+fun <E, A> IOOf<E, A>.fork(ctx: CoroutineContext = IO.dispatchers<Nothing>().default()): IO<Nothing, Fiber<IOPartialOf<E>, A>> =
+  IO {
```
pakoito, line 1215: 👌

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
pakoito: EDIT: yes.

pakoito:
Still approved.

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/fluxk.kt
nomisRev: That bailed with a compile error, `TypeVariable<(U)>`. Not sure what is going wrong there.

nomisRev:
Thanks for the re-review Paco ❤️

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/fluxk.kt
```diff
@@ -157,6 +176,160 @@ interface FluxKAsync :
     fix().continueOn(ctx)
 }
 
+interface FluxKConcurrent : Concurrent<ForFluxK>, FluxKAsync {
+
+  override fun <A> Kind<ForFluxK, A>.fork(coroutineContext: CoroutineContext): FluxK<Fiber<ForFluxK, A>> =
+    fix().fork(coroutineContext)
```
aballano, line 182: nit 

```suggestion
  override fun <A> Kind<ForFluxK, A>.fork(ctx: CoroutineContext): FluxK<Fiber<ForFluxK, A>> =
    fix().fork(ctx)
```

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/fluxk.kt
```diff
@@ -157,6 +176,160 @@ interface FluxKAsync :
...
+    active: AtomicBooleanW,
+    other2: ReactorDisposable,
+    other3: ReactorDisposable,
+    sink: FluxSink<RaceTriple<ForFluxK, A, B, C>>,
+    err: Throwable
+  ): Unit = if (active.getAndSet(false)) {
+    other2.dispose()
+    other3.dispose()
+    sink.error(err)
+  } else Unit
```
aballano, line 316: both onError functions could be local, on the other hand we could merge them into one since we don't really care about the parametric types:

```suggestion
  private inline fun onError(
    active: AtomicBooleanW,
    other2: ReactorDisposable,
    other3: ReactorDisposable?,
    sink: FluxSink<*>,
    err: Throwable
  ): Unit = if (active.getAndSet(false)) {
    other2.dispose()
    other3?.dispose()
    sink.error(err)
  } else Unit
```

(feel free to replace the 2nd disposable nullable type with something fancy if you prefer, a vararg would also be possible as the last argument)

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/extensions/monok.kt
```diff
@@ -118,6 +137,145 @@ interface MonoKAsync : Async<ForMonoK>, MonoKMonadDefer {
...
+    active: AtomicBooleanW,
+    other2: reactor.core.Disposable,
+    sink: MonoSink<RacePair<ForMonoK, A, B>>,
+    err: Throwable
+  ) = if (active.getAndSet(false)) {
+    other2.dispose()
+    sink.error(err)
+  } else Unit
+
+  private fun <A, B, C> onError(
```
aballano, line 252: Same here, this could be grouped as one function

arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
```diff
@@ -91,32 +94,48 @@ data class FluxK<out A>(val flux: Flux<out A>) : FluxKOf<A> {
...
+            Flux.defer { use(a).value() }
+              .doOnCancel { Flux.defer { release(a, ExitCase.Cancelled).value() }.subscribe() }
+              .subscribe({ b ->
+                Flux.defer { release(a, ExitCase.Completed).value() }
+                  .subscribe({ emitter.next(b) }, { e -> emitter.error(e) })
+              }, { e ->
+                Flux.defer { release(a, ExitCase.Error(e)).value() }
+                  .subscribe({ emitter.error(e) }, { e2 -> emitter.error(Platform.composeErrors(e, e2)) })
+              }, {
+                if (!acquireActive.value) emitter.complete()
```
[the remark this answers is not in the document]
aballano: The above meaning that we could potentially check this within a test if it makes sense :D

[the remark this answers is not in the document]
nomisRev: Yes, you're right and it can be worse.

```
val acquire = Flux.just(10, 9, 8, 7, 6, 5, 4, 3, 2, 1).k()
val use = { i: Int -> Flux.timer(i, TimeUnit.SECONDS).map { i }.k() }
val release = { _, _ -> FluxK { println("Release $it") } }

acquire.bracketCase(release, use)
```

- All these timers will start running in parallel since we call `use` immediately in `onNext`
- The first element takes the longest to finish, so when `acquire` completes and `acquireActive` is set to false, and `Flux.timer(1, TimeUnit.SECONDS)` will finish before all other `use` calls are finished.

In other words the `emitter` is closed and the other values from `use` are never observed since they're dropped.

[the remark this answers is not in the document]
aballano: Maybe instead of an AtomicBooleanW we can have an atomic of a list(?) and let onNext's push something (maybe just A) and onComplete push something different (Unit?), so when onNext's `use` finishes we pop onNext's own element and check for the one from onComplete in order to complete 🤔

rachelcarmena:
@nomisRev :raised_hands: the previous comment only shows `Concurrentqueue` because it's the only benchmark included:

```
jmh {
    include = [
            'arrow.benchmarks.Queue'
    ]
    ...
}
```

For instance, for this configuration:

```
jmh {
    include = [
            'arrow.benchmarks.Async',
            'arrow.benchmarks.Queue'
    ]
    ...
}
```

it would appear data for:

* Async.catsIO
* Async.io
* Async.scalazZIO
* Queue.concurrentQueue

**Default threshold** is being used. However, it can be changed in `compareBenchmarksCI` task configuration: https://47degrees.github.io/hood/gradle/ci/

nomisRev:
Closing due to getting deprecated.

Issue #145, opened 2020-04-10
BIO: Add APIs to work with Either


pakoito:
```
fun <E, EE : E, A> IO<E, Either<E, A>>.flattenEither(): IO<E, A> =
  flatMap { it.fold(::IO.raiseError, ::IO.just) }
```

```
fun <E, A> effectEither(f: suspend () -> Either<E, A>): IO<E, A> =
  effect(f).flattenEither()
```

```
fun <E, EE: E, A, B> IO<E, A>.mapEither(f: (A) -> Either<EE, B>): IO<EE, B> =
  map(f).flattenEither()
```

```
fun <E, EE: E, A, B> IO<E, A>.effectMapEither(f: suspend (A) -> Either<EE, B>): IO<EE, B> =
  effectMap(f).flattenEither()
```

raulraja:
Can effect and map for Either also use the same names as for the non either shape. Would this be more efficient if it's treated as a new IO node in the run loop like Effect?

pakoito:
They cannot because the compiler cannot tell `(A) -> B` and `(A) -> Either<E, B>` apart at the callsites.

raulraja:
If it's because of JVM Clash we could use @JvmName

Hiosdra:
Shouldn't `mapEither` be named `flatMapEither`? I think `mapEither` should be more of:
```kotlin
fun <E, A> IO<E, A>.mapEither(f: (A) -> Left<E>): IO<E, A>
```
, but it does not make sense to include it in arrow in my opinon.

However, `flatMapEither` would be great 😀
---
Is any chance to make `flatMapKind` (not this naming 😝 ) which will be something of:
`Kind<F, A>.flatMapKind(A -> Kind<G, A>)`? I think of some abstraction above IO and Either. 
It could take `A -> IO`, `A -> Either`, `A -> Option` etc and will be available on any of this types, so in para-haskell way:
```
IO -> A -> IO
IO -> A -> Either
IO -> A -> Option
Either -> A -> IO
Either -> A -> Either
Either -> A -> Option
Option -> A -> IO
Option -> A -> Either
Option -> A -> Option
```

I see challange in making it work with both `Kind` and `Kind2`, but maybe it is possible?

pakoito:
`Either` is specialised here because it'll the the most frequent use. I can see the API working for all `ApplicativeError`, although the ergonomy will be meh.

Hiosdra:
[...]

PR #157, opened 2020-05-04
Add IO APIs to work with Either

pablisco:
Adds extra support for Either in IO as mentioned in #145

### DIFF
--- a/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
+++ b/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
@@ -1095,7 +1095,7 @@ fun <E, A, B> IO<E, A>.mapEither(f: (A) -> EitherOf<E, B>): IO<E, B> =
  *   suspend fun Int.increment() = Either.right(this + 1)
  *   val result =
  *   //sampleStart
- *   IO.just(1).mapEither { it.increment() }
+ *   IO.just(1).effectMapEither { it.increment() }
  *   //sampleEnd
  *   println(result.unsafeRunSyncEither())
  * }
--- a/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
+++ b/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
@@ -1061,7 +1061,7 @@ fun <E, A> IO<E, EitherOf<E, A>>.flatten(): IO<E, A> =
  * }
  * ```
  */
-fun <E, A, B, E2 : E> IO<E2, A>.mapEither(f: (A) -> EitherOf<E, B>): IO<E, B> =
+fun <E, A, B> IO<E, A>.mapEither(f: (A) -> EitherOf<E, B>): IO<E, B> =
   map(f).flatten()
 
 /**
@@ -1079,7 +1079,7 @@ fun <E, A, B, E2 : E> IO<E2, A>.mapEither(f: (A) -> EitherOf<E, B>): IO<E, B> =
  * }
  * ```
  */
-fun <E, A, B, E2 : E> IO<E2, A>.effectMapEither(f: suspend (A) -> EitherOf<E, B>): IO<E, B> =
+fun <E, A, B> IO<E, A>.effectMapEither(f: suspend (A) -> EitherOf<E, B>): IO<E, B> =
   effectMap(f).flatten()
 
 /**
--- a/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
+++ b/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
@@ -16,6 +16,7 @@ import arrow.core.fix
 import arrow.fx.IO.Companion.async
 import arrow.fx.IO.Companion.effect
 import arrow.fx.IO.Companion.just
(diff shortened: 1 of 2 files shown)


rachelcarmena:
@pablisco :raised_hands: thanks!! I've just pushed an empty commit to re-run checks after [the last incident from GitHub](https://www.githubstatus.com/incidents/ljygds069np9).

pakoito:
To prod it goes!

arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -994,6 +1046,64 @@ fun <E, A, B, E2 : E> IOOf<E, A>.flatMap(f: (A) -> IOOf<E2, B>): IO<E2, B> =
...
+ *   fun Int.increment() = Either.right(this + 1)
+ *   val result =
+ *   //sampleStart
+ *   IO.just(1).mapEither { it.increment() }
+ *   //sampleEnd
+ *   println(result.unsafeRunSyncEither())
+ * }
+ * ```
+ */
+fun <E, A, B> IO<E, A>.mapEither(f: (A) -> EitherOf<E, B>): IO<E, B> =
```
aballano, line 1086: If I'm not mistaken, this function should already be covered by `effectMapEither` because you can still pass a non-suspend function to that one. So I'd personally remove this one, no? cc @pakoito @nomisRev

arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -994,6 +1046,64 @@ fun <E, A, B, E2 : E> IOOf<E, A>.flatMap(f: (A) -> IOOf<E2, B>): IO<E2, B> =
...
+/**
+ * Transform, as a suspend effect, the value of an [IO] into an [Either] and consequently flatten into an [IO]
+ *
+ * @return success when [f] results in [Either.Right] or error when [f] results in [Either.Left]
+ *
+ * fun main(args: Array<String>) {
+ *   suspend fun Int.increment() = Either.right(this + 1)
+ *   val result =
+ *   //sampleStart
+ *   IO.just(1).mapEither { it.increment() }
```
aballano, line 1098:
```suggestion
 *   IO.just(1).effectMapEither { it.increment() }
```

arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -994,6 +1046,64 @@ fun <E, A, B, E2 : E> IOOf<E, A>.flatMap(f: (A) -> IOOf<E2, B>): IO<E2, B> =
...
+ *
+ *  * ```kotlin:ank:playground
+ * import arrow.fx.IO
+ * import arrow.fx.unsafeRunSync
+ * import arrow.core.Either
+ *
+ * fun main(args: Array<String>) {
+ *   val result =
+ *   //sampleStart
+ *   IO.just(Either.right("Hello")).flatten().flatMap { IO { "$it World" } }
```
aballano, line 1062:
```suggestion
 *   IO.just(Either.right("Hello")).flattenEither().flatMap { IO { "$it World" } }
```

arrow-fx/src/main/kotlin/arrow/fx/IO.kt
pablisco: It does work, however, we may not force a suspension unless necessary. The bytecode is quite different:
For:
```kotlin
val run = IO.just(1).mapEither { Either.right(it + 1) }
```
it generates the equivalent to:
```Java
run = IOKt.mapEither(IO.Companion.just(1), (Function1)null.INSTANCE);
```
On the other hand, if we use suspension:
```kotlin
val runAsync = IO.just(1).effectMapEither { Either.right(it + 1) }
```
it generates the equivalent to:
```java
     runAsync = IOKt.effectMapEither(IO.Companion.just(1), (Function2)(new Function2((Continuation)null) {
         private int p$0;
         int label;

         @Nullable
         public final Object invokeSuspend(@NotNull Object $result) {
            Object var3 = IntrinsicsKt.getCOROUTINE_SUSPENDED();
            switch(this.label) {
            case 0:
               ResultKt.throwOnFailure($result);
               int it = this.p$0;
               return Either.Companion.right(Boxing.boxInt(it + 1));
            default:
               throw new IllegalStateException("call to 'resume' before 'invoke' with coroutine");
            }
         }

         @NotNull
         public final Continuation create(@Nullable Object value, @NotNull Continuation completion) {
            Intrinsics.checkParameterIsNotNull(completion, "completion");
            Function2 var3 = new <anonymous constructor>(completion);
            ((Number)value).intValue();
            var3.p$0 = ((Number)value).intValue();
            return var3;
         }

         public final Object invoke(Object var1, Object var2) {
            return ((<undefinedtype>)this.create(var1, (Continuation)var2)).invokeSuspend(Unit.INSTANCE);
         }
      }));
```

pakoito: Yep, it's a perf thing. We want `map` separate from `effectMap`.

PR #158, opened 2020-05-08
Android Lifecycle integration

### LINKED ISSUE #152: ["Request"] Android Lifecycle
What would you like to see?

An Arrow-Fx Android lifecycle integration for `IO`, similar to the KotlinX Coroutines integration module.

It should expose the same functionality as KotlinX Coroutines but based on `LifecycleOwner` instead of `CoroutineScope`.

See [AutoDispose](https://github.com/uber/AutoDispose) for an RxJava implementation.

aballano:
Same API as the Kotlinx coroutines integration that provides auto cancelation for LifecycleOwners (Activities, Fragments, etc.). This part you can already review :)

I also added a `deferUntilActive` function that allows devs to defer the emission of an IO until the LifecycleOwner is in the Create state, so it should behave like a LiveData.

TODO:
* ~Figure out if it's possible to move the root build.gradle's added config for android to the specific module~ Thanks @rachelcarmena!
* Enable publish and submodule config on the module (currently failing)
* ~We might need to add the android sdk to the CI somehow to test this module~
* ~Finish the tests for `deferUntilActive`~ removed for now

Fixes #152


arrow-fx-android/src/main/kotlin/arrow/integrations/android/extensions.kt
```diff
@@ -0,0 +1,43 @@
...
+ * Unsafely run an [IO] and receive the values in a callback [cb] while participating in structured concurrency.
+ * Equivalent of [IO.unsafeRunAsyncCancellableEither] but with its cancellation token wired to [LifecycleOwner].
+ *
+ * @see [IO.unsafeRunAsyncCancellableEither] for a version that returns the cancellation token instead.
+ */
+fun <E, A> IOOf<E, A>.unsafeRunScoped(
+  owner: LifecycleOwner,
+  cb: (IOResult<E, A>) -> Unit
+) {
+  if (owner.lifecycle.currentState.isAtLeast(State.CREATED)) {
```
pakoito, line 31: Why is this initial state required?

aballano: Otherwise we potentially might launch an IO after the onDestroy phase and therefore might never be cancelled

arrow-fx-android/gradle.properties
```diff
@@ -0,0 +1,5 @@
+# Maven publishing configuration
+POM_NAME=Arrow-Fx-Android
+POM_ARTIFACT_ID=arrow-fx-android
+POM_PACKAGING=aar
```
rachelcarmena, line 4:
```suggestion
POM_PACKAGING=jar
```

aballano: @rachelcarmena I personally choose aar since android libs work this way additionally, so it's like a requirement to the project implementing to be an android lib/app, does it make sense? Otherwise we can revert to jar since it's compatible anyway 🤷

rachelcarmena: Oh, ok! I thought it was a typo :joy:

aballano:
@nomisRev I'm thinking on ditching out the `deferUntilActive` fun since it might not be that simple: what livedata does is have an inner count of auto-removal-subscribers to track if it's active, which doesn't make sense for IO but for Stream I presume. Also with the other changes coming to suspend I think we could safely revisit the `launchWhenXX` syntax then.

rachelcarmena:
@aballano , I fixed the first TODO item in #162.

About the second one:

* Enable publish and submodule config on the module (currently failing)

They are failing because they are based on Java plugin and it's not compatible with the Android plugins. It seems Android modules will need a different configuration for testing and publication. Let's research it! :raised_hands:

nomisRev:
@rachelcarmena yes, releasing Android involves a different setup. @JorgeCastilloPrz has some released Android libraries, he can help us there.

JorgeCastilloPrz:
In [this repo](https://github.com/JorgeCastilloPrz/hiroaki) you can see how [the deploy script for the android module](https://github.com/JorgeCastilloPrz/hiroaki/blob/master/hiroaki-android/gradle-mvn-push.gradle) is slightly different from [the pure kotlin one](https://github.com/JorgeCastilloPrz/hiroaki/blob/master/hiroaki-core/gradle-mvn-push.gradle).

arrow-fx-android/src/main/kotlin/arrow/integrations/android/extensions.kt
JorgeCastilloPrz: Doesn't it feel a bit confusing from the user perspective? If the condition is not matched the dev doesn't get any error but the job is not run.

If the condition is not matched we could throw, if we wanna be strict (this is already an unsafe run), or alternatively, we could install a new lifecycle observer in that case, so it runs the job automatically if the lifecycle reaches de CREATED state afterwards. (Note this could get called earlier in the lifecycle).

I've not dug into the implementation yet, but I'd say something like that is what the [Android lifecycle aware coroutines](https://developer.android.com/topic/libraries/architecture/coroutines#suspend) do under the hood.

aballano: > (Note this could get called earlier in the lifecycle)

I don't think so, at least from what I understood here https://developer.android.com/reference/androidx/lifecycle/Lifecycle.State#summary the CREATED state is the minimum active state you can have

> or alternatively, we could install a new lifecycle observer in that case, so it runs the job automatically if the lifecycle reaches de CREATED state afterwards. 

As far as I'm concerned that wouldn't work because of this (taken from the `launchWhenXXX` fun)

image

(where `handleDestroy` is simply cancelling the parent and finishing)

> I've not dug into the implementation yet, but I'd say something like that is what the Android lifecycle aware coroutines do under the hood with those launchWhenXXXX methods.

Note that we're in the `unsafeRunScoped` function which is different from the `launchWhenXXX`, the latter can be easily implementing by simply flatmapping an `ioOfEvent` to the IO we want to start at that state:

```
IO.async { cb ->	
    owner.lifecycle.addObserver(object : LifecycleEventObserver {	
      override fun onStateChanged(source: LifecycleOwner, event: Event) {	
        if (event == eventTarget) {	
          source.lifecycle.removeObserver(this)	
          cb(IOResult.Success(Unit))	
        }	
      }	
    })	
  }
```

arrow-fx-android/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+    id "org.jlleitschuh.gradle.ktlint"
+    id "ru.vyarus.animalsniffer"
+}
+
+repositories {
+    google()
+    jcenter()
+}
+
+//apply from: "$SUBPROJECT_CONF"
```
danimontoya, line 24: this should be uncommented?

arrow-fx-android/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+apply from: "$DOC_CONF"
+//apply from: "$PUBLISH_CONF"
```
danimontoya, line 26: and this one too?

aballano: Yes, both are part of the leftover point in the description: `Enable publish and submodule config on the module (currently failing)`

arrow-fx-android/build.gradle
```diff
@@ -0,0 +1,54 @@
+buildscript {
+    repositories {
+        google()
+    }
+    dependencies {
+        classpath("com.android.tools.build:gradle:3.6.0")
```
JorgeCastilloPrz, line 6: This can be 4.0.0 now. https://twitter.com/AndroidDev/status/1266054894043058176?s=19

Issue #76, opened 2020-03-03
Include benchmarks on pull requests


rachelcarmena:
Benchmarks are just being compiled on every pull request so far.

It would be interesting to include a comparison of benchmarks results between `master` branch and the pull request branch.

rachelcarmena:
It's solved via #159

PR #159, opened 2020-05-08
Solve #76. Add benchmark comparison

rachelcarmena:
This PR adds benchmark comparison via [Hood](https://47degrees.github.io/hood/) so a comment with the result will be added in every pull request on this repository. 

**See the comment below** as an example. 

That comment will be **updated** with new commits:

Screenshot from 2020-05-08 19-27-56

In this case there isn't difference between values because this pull request is just adding the configuration and it's not changing Arrow Fx library.

It only shows `Concurrentqueue` because it's the only benchmark included:

```
jmh {
    include = [
            'arrow.benchmarks.Queue'
    ]
    ...
}
```

For instance, for this configuration:

```
jmh {
    include = [
            'arrow.benchmarks.Async',
            'arrow.benchmarks.Queue'
    ]
    ...
}
```

it would appear data for:

* Async.catsIO
* Async.io
* Async.scalazZIO
* Queue.concurrentQueue

**Default threshold** is being used. However, it can be changed in `compareBenchmarksCI` task configuration: https://47degrees.github.io/hood/gradle/ci/

For the rest of open pull requests, the comment with the results will appear when merging this pull request and then updating those pull requests with master branch.

Thanks to @AdrianRaFo for providing me the guide about it :raised_hands:

### DIFF
--- a/.github/workflows/build_arrow-fx.yml
+++ b/.github/workflows/build_arrow-fx.yml
@@ -31,13 +31,14 @@ jobs:
     - name: Run benchmark for master branch
       run: |
         git checkout master
+        # TODO: When this file is located on master branch
+        # ./gradlew :arrow-benchmarks-fx:jmhForMasterBranch
         ./gradlew :arrow-benchmarks-fx:jmh
         cp arrow-benchmarks-fx/build/reports/benchmarks.json master-branch.json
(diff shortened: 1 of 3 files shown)


rachelcarmena:
Thanks @aballano :raised_hands:

PR #166, opened 2020-05-29
Fix Either.fx suspend usage

aballano:
Fixes https://github.com/arrow-kt/arrow-core/issues/135


aballano:
Thank you fast @pakoito 😘

Issue #174, opened 2020-06-08
["BUG"] Schedule.never


nomisRev:
What version are you currently using?

`0.10.5`

What would you like to see?

`Schedule.never` should wait forever like all other `F` `never` operators.

[Fixed Arrow Fx Coroutines source](https://github.com/arrow-kt/arrow-fx/blob/369aea0115a8368ab3fef2a48a00ada76fa1c0df/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt#L622)

[Broken Arrow Fx source](https://github.com/arrow-kt/arrow-fx/blob/master/arrow-fx/src/main/kotlin/arrow/fx/Schedule.kt#L713)

aballano:
@nomisRev the problem here with never is that in the Coroutines impl you're able to pass a never like `invoke(suspend { arrow.fx.coroutines.never<Unit>() })` but AFAIK we cannot do that right now with Monad but rather Async, no?

nomisRev:
We have a constraint of `MonadDefer` for `jitterred`. We should do the same here for `never`.

Issue #173, opened 2020-06-08
["BUG"] Schedule.recurs


nomisRev:
What version are you currently using?

`0.10.5`

What would you like to see?

`Schedule.recurs` incorrectly runs.

[Fixed Arrow Fx Coroutines source](https://github.com/arrow-kt/arrow-fx/blob/sv-arrow-fx-coroutines/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt#L608)

[Broken Arrow Fx source](https://github.com/arrow-kt/arrow-fx/blob/master/arrow-fx/src/main/kotlin/arrow/fx/Schedule.kt#L701)

Issue #172, opened 2020-06-08
["BUG"] repeat `Schedule`


nomisRev:
What version are you currently using?

`0.10.5`

What would you like to see?

`repeat` `Schedule` behaves incorrectly. See: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1591017049041700

See Arrow Fx Coroutines Schedule implementation for fix.

PR #184, opened 2020-06-11
Fixes to Schedule

aballano:
* Fix schedule recurs #173 returning the number of total iterations, instead of repetitions
* Fix never #174 that should never repeat and basically hang the program
* Fix repeat not waiting for delay after first emission #172 
* Also removed Alternative and MonoidK instances as it might not have sense to provide Schedule.never() as MonoidK.empty()

### DIFF
--- a/arrow-fx/src/main/kotlin/arrow/fx/Schedule.kt
+++ b/arrow-fx/src/main/kotlin/arrow/fx/Schedule.kt
@@ -401,7 +401,7 @@ sealed class Schedule<F, Input, Output> : ScheduleOf<F, Input, Output> {
 
   /**
    * Add random jitter to a schedule.
-   * The argument [genRand] is supposed to be a computation with when run returns
+   * The argument [genRand] is supposed to be a computation that returns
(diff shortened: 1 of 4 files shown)


arrow-fx/src/main/kotlin/arrow/fx/Schedule.kt
```diff
@@ -708,10 +714,13 @@ sealed class Schedule<F, Input, Output> : ScheduleOf<F, Input, Output> {
 
     /**
      * Create a schedule that never retries.
```
nomisRev, line 716: We should make clear here that it'll never return from `retry` or `repeat` until canceled.

This is useful in similar situations as `never` where you need to explicitly suspend until canceled or interrupted. You could compose it at the end of a `Schedule` using `andThen` to suspend after a `Schedule` finishes.

aballano: added a note "Note that this will hang a program if used as a repeat/retry schedule unless cancelled."

PR #189, opened 2020-06-17
Make `IQueue` size `O(1)`

nomisRev:
Small improvement to `IQueue`.
It uses a `property` size to track the size of the `Queue` instead of relying the `List` implementation.


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/IQueue.kt
```diff
@@ -31,13 +38,10 @@ internal class IQueue<A> internal constructor(
       else -> null
     }
 
-  fun tail(): IQueue<A> =
```
JorgeCastilloPrz, line 34: 👍

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/IQueue.kt
```diff
@@ -47,52 +51,68 @@ internal class IQueue<A> internal constructor(
...
+
+  fun filterNot(p: (A) -> Boolean): IQueue<A> {
+    val newIn = listIn.filterNot(p)
+    val newOut = listOut.filterNot(p)
+    return IQueue(newIn, newOut, newIn.size + newOut.size)
+  }
 
-  fun filterNot(p: (A) -> Boolean): IQueue<A> =
-    IQueue(listIn.filterNot(p), listOut.filterNot(p))
+  override fun equals(other: Any?): Boolean =
```
JorgeCastilloPrz, line 95: Is there a reason for IQueue to not be a data class? Given its a class defined by its properties

nomisRev: No good point!
It can definitely be a data class, I'd still overwrite the `toString` to the new one tho.

JorgeCastilloPrz: That'd make sense to me.

nomisRev:
Thanks for the review @JorgeCastilloPrz 🙌

PR #190, opened 2020-06-17
Reduce test stress

nomisRev:
Reduced the stress of our test suite by 10x by reducing the iterations to 100 instead of 1000.

This has shaved off ~ 7minutes on my local machine, while still covering all `edgeCases`.
There are two implementations here based on `Kotest` in Arrow Fx Coroutines, and `KotlinTest` in the rest.
We should push this change down to `Arrow Core Test`, so we can also configure `iterations` for law testing.

### Proposed follow up work

Ideally, we'd decouple property testing from `Kotest` so:

```diff
-fun <A> Monoid<A>.monoidLeftIdentity(GEN: Gen<A>, EQ: Eq<A>): Unit =
+fun <A> Monoid<A>.monoidLeftIdentity(EQ: Eq<A>): Boolean =
-  forAll(GEN) { a ->
    (empty().combine(a)).equalUnderTheLaw(a, EQ)
-  }
```

This allows for more flexibility as you could plug it into both `KotlinTest` & `Kotest` and that simply becomes a integration instead of hard-coupled as it is today.

This also decouples the property testing from the law definition, and would allow us to easily control test strategy by the caller.


JorgeCastilloPrz:
I know this is a hard to answer question but, how many iterations are okay to ensure code integrity? Could 100 iterations be too low to test an acceptable amount of edge cases?

nomisRev:
@JorgeCastilloPrz Kotest checks that you have enough iterations to cover all `edgeCases`, so we're guaranteed to cover all `edgeCases` + `iterations - edgeCases.size` random values.

https://github.com/kotest/kotest/blob/master/kotest-property/src/commonMain/kotlin/io/kotest/property/internal/proptest.kt#L17

I just double checked for KotlinTest, and it doesn't have the same assurance but we need to refactor to Kotest asap since we're a version behind now.

arrow-fx-test/src/main/kotlin/arrow/fx/test/laws/InvariantLaws.kt
```diff
@@ -1,44 +0,0 @@
...
-import arrow.core.test.generators.GenK
-import arrow.core.test.generators.functionAToB
-import arrow.core.test.laws.Law
-import arrow.typeclasses.Eq
-import arrow.typeclasses.EqK
-import arrow.typeclasses.Invariant
-import io.kotlintest.properties.Gen
-import io.kotlintest.properties.forAll
-
-object InvariantLaws {
```
aballano, line 16: Why is this removed?

nomisRev: It's duplicated with Arrow-Core-Test.

nomisRev:
@aballano it's definitely combine-able with arrow-kt/arrow-core#105, but I feel like we should come up with a strategy that also incorporates upgrading to kotest.

PR #198, opened 2020-06-26
Use of KOTEST_VERSION property

rachelcarmena:
In order to help to the progressive use of Kotest.

**Note**: it corresponds to `4.1.0` instead of `4.0.6`. Let's see if it works...

### DIFF
--- a/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/predef-test.kt
+++ b/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/predef-test.kt
@@ -111,13 +111,9 @@ fun <L, R> Arb.Companion.either(left: Arb<L>, right: Arb<R>): Arb<Either<L, R>> 
   return Arb.choice(failure, success)
 }
 
-fun Arb.Companion.intRange(min: Int = 0, max: Int = 1000): Arb<IntRange> =
-  Arb.int(min, max).flatMap { a ->
-    Arb.int(min, max).map { b ->
-      val first = min(a, b)
-      val last = max(a, b)
-      first..last
-    }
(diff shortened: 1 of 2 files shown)


arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/predef-test.kt
```diff
@@ -111,13 +110,9 @@ fun <L, R> Arb.Companion.either(left: Arb<L>, right: Arb<R>): Arb<Either<L, R>>
...
 }
 
-fun Arb.Companion.intRange(min: Int = 0, max: Int = 1000): Arb<IntRange> =
-  Arb.int(min, max).flatMap { a ->
-    Arb.int(min, max).map { b ->
-      val first = min(a, b)
-      val last = max(a, b)
-      first..last
-    }
+fun Arb.Companion.intRange(min: Int = Int.MIN_VALUE, max: Int = Int.MAX_VALUE): Arb<IntRange> =
```
rachelcarmena, line 113: @nomisRev , thank you so much for making it work for Kotest `4.1.0` :tada: Just a small question, could this change impact on current tests? I mean, the max value changed from 1000 to 2147483647. Thanks in advance!!

rachelcarmena: And the min value also changes (from 0 to -2147483648).

nomisRev: Hey @rachelcarmena,
The reason for this change was to make the generator more accurate.
The only current usage from this `Arb` explicitly defines `min` & `max`, so it doesn't impact the current suite.

rachelcarmena: Great, thank you so much Simon!! :raised_hands:

PR #199, opened 2020-06-26
Adds Stream.callback

aballano:
Also fixes a couple of nits found in docs


JorgeCastilloPrz:
Some Ktlint warnings

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackStreamTest.kt
```diff
@@ -0,0 +1,162 @@
...
+    Stream.callbackStream<Int> {
+      emit(listOf(1, 2, 3))
+    }
+      .take(3)
+      .compile()
+      .toList() shouldBe listOf(1, 2, 3)
+  }
+
+  "emits chunks" {
+    checkAll(Arb.chunk(Arb.int()).filter { it.isNotEmpty() }, Arb.chunk(Arb.int()).filter { it.isNotEmpty() }) { ch, ch2 ->
```
nomisRev, line 67: This test shows that emitting `Chunk.empty` breaks this constructor. This should be fixed, or we should create a ticket for it if we want to unblock @JorgeCastilloPrz by merging.

This can be fixed by without using `null` or `Option`, by changing to `Queue.unbounded<Any?>`.
And using reference checking/casting + `object End`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,124 @@
...
+ *     .let(::println) //[1, 2, 3, 4]
+ * //sampleEnd
+ * ```
+ *
+ * Note that if neither `end()`, `emit(Chunk.empty())` nor other limit operators such as `take(N)` are called,
+ * then the Stream will never end.
+ */
+// @OptIn(ExperimentalTypeInference::class) in 1.3.70
+@UseExperimental(ExperimentalTypeInference::class)
+fun <A> Stream.Companion.asyncCancellable(@BuilderInference f: suspend EmitterSyntax<A>.() -> CancelToken): Stream<A> =
```
nomisRev, line 77:
```suggestion
fun <A> Stream.Companion.cancellable(@BuilderInference f: suspend EmitterSyntax<A>.() -> CancelToken): Stream<A> =
```

We should rename this to be `cancellable` such as that it follows the same familira `Async/Concurrent` naming.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,124 @@
...
+  effect {
+    val q = Queue.unbounded<Any?>()
+    val error = UnsafePromise<Throwable>()
+
+    val cancel = emitterCallback(f) { value ->
+      suspend {
+        q.enqueue1(value)
+        // TODO shall we consider emitting from different contexts? Might serve as an observeOn in RxJava
+      }.startCoroutine(Continuation(EmptyCoroutineContext) { r -> r.fold({ Unit }, { e -> error.complete(Result.success(e)) }) })
```
nomisRev, line 86:
```suggestion
      }.startCoroutine(Continuation(ComputationPool) { r -> r.fold({ Unit }, { e -> error.complete(Result.success(e)) }) })
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,124 @@
...
+@UseExperimental(ExperimentalTypeInference::class)
+fun <A> Stream.Companion.asyncCancellable(@BuilderInference f: suspend EmitterSyntax<A>.() -> CancelToken): Stream<A> =
+  effect {
+    val q = Queue.unbounded<Any?>()
+    val error = UnsafePromise<Throwable>()
+
+    val cancel = emitterCallback(f) { value ->
+      suspend {
+        q.enqueue1(value)
+        // TODO shall we consider emitting from different contexts? Might serve as an observeOn in RxJava
```
nomisRev, line 85: The context where is emitted from is never seen by anyone.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,124 @@
...
+import arrow.fx.coroutines.ExitCase
+import arrow.fx.coroutines.UnsafePromise
+import arrow.fx.coroutines.andThen
+import arrow.fx.coroutines.stream.concurrent.Queue
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.startCoroutine
+import kotlin.experimental.ExperimentalTypeInference
+
+private object END
```
nomisRev, line 14: 👍

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/StreamTest.kt
```diff
@@ -768,15 +768,15 @@ class StreamTest : StreamSpec(spec = {
...
-                .flatMap { loop(acc + uncons1.head, uncons1.tail) }
+                .append { loop(acc + uncons1.head, uncons1.tail) }
             }
           }
 
         loop(0, it.asPull()).stream()
       }
 
       Stream.iterate(0, Int::inc)
-        .flatMap { Stream(it).delayBy(20.milliseconds) }
```
nomisRev, line 779: We've discussed `50.milliseconds` as a minimum delay in another PR. Why the decrease here?

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackStreamTest.kt
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  "emits iterable" {
+    checkAll(Arb.list(Arb.int())) { list ->
+      Stream.async<Int> {
+        emit(list)
+        end()
+      }
+        .compile()
+        .toList() shouldBe list
```
nomisRev, line 68: Should be chunked.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackStreamTest.kt
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  "emits varargs" {
+    checkAll(Arb.list(Arb.int())) { list ->
+      Stream.async {
+        emit(*list.toTypedArray())
+        end()
+      }
+        .compile()
+        .toList() shouldBe list
```
nomisRev, line 57: Should be chunked.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/StreamTest.kt
```diff
@@ -768,15 +768,15 @@ class StreamTest : StreamSpec(spec = {
             when (uncons1) {
               null -> Pull.output1(acc)
               else -> Pull.output1(uncons1.head)
-                .flatMap { loop(acc + uncons1.head, uncons1.tail) }
+                .append { loop(acc + uncons1.head, uncons1.tail) }
```
nomisRev, line 771: This doesn't change anything.
`flatMap == append` if the stream is doesn't interrupt/fail.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
nomisRev: Yes, that's correct.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * Note that if neither `end()` nor other limit operators such as `take(N)` are called,
+ * then the Stream will never end.
+ *
+ * Cancellation or errors might happen any time during emission, so it's recommended to call `onCancel` as early as
+ * possible, otherwise it's not guaranteed to be called.
+ */
+// @OptIn(ExperimentalTypeInference::class) in 1.3.70
+@UseExperimental(ExperimentalTypeInference::class)
+fun <A> Stream.Companion.async(@BuilderInference f: suspend EmitterSyntax<A>.() -> Unit): Stream<A> =
```
aballano, line 54: To be: Stream.Companion.callback

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * Note that if neither `end()` nor other limit operators such as `take(N)` are called,
+ * then the Stream will never end.
+ *
+ * Cancellation or errors might happen any time during emission, so it's recommended to call `onCancel` as early as
+ * possible, otherwise it's not guaranteed to be called.
+ */
+// @OptIn(ExperimentalTypeInference::class) in 1.3.70
+@UseExperimental(ExperimentalTypeInference::class)
+fun <A> Stream.Companion.callback(@BuilderInference f: suspend EmitterSyntax<A>.() -> Unit): Stream<A> =
```
nomisRev, line 54: We should fix the inconsititecy between this signature and [`cancellable`](https://github.com/arrow-kt/arrow-fx/blob/e6a8618fbf2450a4084d105bf88ba563c0cc836b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/builders.kt#L7) & [`cancellableF`](https://github.com/arrow-kt/arrow-fx/blob/e6a8618fbf2450a4084d105bf88ba563c0cc836b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/builders.kt#L27) from `IO` and `Arrow Fx Coroutines`.

`suspend EmitterSyntax<A>.() -> Unit` should become `suspend EmitterSyntax<A>.() -> CancelToken`, that guarantees that when a `CancelToken` is returned it's run. When no token is returned, it can also not be run. I.e. when running blocking code inside `cancellable`. It however enforces the caller of `callback` to return a `CancelToken`.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackTest.kt
```diff
@@ -0,0 +1,235 @@
...
+
+      parTupledN({ latch.get() }, { sleep(20.milliseconds) })
+
+      f.cancel()
+
+      effect.get() shouldBe true
+    }
+  }
+
+  "can cancel never" {
```
aballano, line 187: We would need to refactor this test somehow, any idea? cc @nomisRev

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackTest.kt
```diff
@@ -0,0 +1,235 @@
...
+import io.kotest.matchers.shouldBe
+import io.kotest.property.Arb
+import io.kotest.property.arbitrary.int
+import io.kotest.property.arbitrary.list
+import io.kotest.property.arbitrary.map
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.startCoroutine
+
+class CallbackTest : StreamSpec(iterations = 250, spec = {
```
pablisco, line 34: Should we add a test for when we emit elements after calling `end()`?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,143 @@
...
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.startCoroutine
+import kotlin.experimental.ExperimentalTypeInference
+
+private object END
+
+interface EmitterSyntax<A> {
+  fun emit(a: A): Unit
+  fun emit(chunk: Chunk<A>): Unit
+  fun emit(iterable: Iterable<A>): Unit
```
pablisco, line 22: Should we include an option for emitting a `Stream<A>`?

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackTest.kt
pablisco: Another potential test case could be if we emit from a separate coroutine. And also if we emit from a separate coroutine after the stream has ended.

nomisRev: If you  emit after `end` it's offloaded in the `Queue` but never emitted downstream. Would you expect an exception here?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
```diff
@@ -1,8 +1,11 @@
 package arrow.fx.coroutines
 
-inline infix fun <A, B, C> ((A) -> B).andThen(crossinline f: (B) -> C): (A) -> C =
+internal inline infix fun <A, B, C> ((A) -> B).andThen(crossinline f: (B) -> C): (A) -> C =
   { a -> f(this(a)) }
 
+internal inline infix fun <A, B, C> (suspend (A) -> B).andThen(crossinline f: suspend (B) -> C): suspend (A) -> C =
```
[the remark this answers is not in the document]
nomisRev: These could go into Arrow-Core, we currently have them in a `syntax` module I believe. We should open tickets for that.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,143 @@
...
+import arrow.fx.coroutines.UnsafePromise
+import arrow.fx.coroutines.andThen
+import arrow.fx.coroutines.guaranteeCase
+import arrow.fx.coroutines.stream.concurrent.Queue
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.startCoroutine
+import kotlin.experimental.ExperimentalTypeInference
+
+private object END
```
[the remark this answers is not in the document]
nomisRev: Very good point! Implemented this suggestion 👍

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,123 @@
...
+
+interface EmitterSyntax<A> {
+  fun onCancel(cancelF: suspend () -> Unit): Unit
+  fun emit(a: A): Unit
+  fun emit(chunk: Chunk<A>): Unit
+  fun emit(iterable: Iterable<A>): Unit
```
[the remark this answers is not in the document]
nomisRev: Arrow Fx Coroutine Stream works differnt from KotlinX Coroutine Flow, both are `suspend` pull based, but Arrow Fx's Stream works `Chunk` based. So `emit(a: A)` is actually the special case here, since that emits a singleton `Chunk`.

I don't think we should be concerned about matching other libraries API :D

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ParJoin.kt
```diff
@@ -146,9 +146,17 @@ suspend fun <O> Stream<Stream<O>>.runOuter(
...
 // and then collects result of the stream (outer + inner) execution
-suspend fun signalResult(done: SignallingAtomic<Option<Option<Throwable>>>): Unit =
+internal suspend fun signalResult(done: SignallingAtomic<Option<Option<Throwable>>>): Unit =
   done.get().flatten().fold({ Unit }, { throw it })
 
+/**
+ * Merges both Streams into an Stream of A and B represented by Either<A, B>.
+ * This operation is equivalent to a normal merge but for different types.
+ */
+fun <A, B> Stream<A>.either(other: Stream<B>): Stream<Either<A, B>> =
```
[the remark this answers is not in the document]
nomisRev: What would be the rationale behind calling `fun <A> Stream<A>.merge(other: Stream<A>): Stream<A>` the same way?

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackTest.kt
nomisRev: Hey @aballano,
Sorry for the long wait :/ I aligned the semantics of `Stream.cancellable` to `cancellable` as discussed, and also tried to update the documentation and examples to clarify how these functions work.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackTest.kt
pablisco: Ideally, in the future we can have restrictions by the compiler. I started a chat about this on the KotlinLang Slack.

I guess throwing an exception at the calling site may make sense... As consumers have probably removed their subscription after the end signal. On other coroutine APIs they do emit and offer to differentiate if the consumer cares or not about the success of the push.

We could have a 3rd one that returns a result or either... Maybe we can have this as a separate ticket

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ParJoin.kt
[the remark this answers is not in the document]
pablisco: Ah yeah, this is more a concatenation operation. I guess the name didn't give me enough information. Also, the first word in the docs is "Merge" 😅 
Maybe we can change it to concat instead? (Maybe with the plus operator too)

When we have unions in the future could even generalise it to concatenate streams for more than two types 🤞

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
pablisco: Could potentially emit with two strategies:
- emit: waits for the end of the stream before continuing
- emitPar: pushes the stream up in parallel

The end result would be similar to concatenate and merge operations respectively

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ParJoin.kt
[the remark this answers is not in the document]
pablisco: Nevermind. I was reading it on my phone and I just noticed the name of the class as parJoin 🤦 
I guess the name `join` or `parJoin` may be more fitting? Not a blocker though :)

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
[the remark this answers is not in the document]
pablisco: Moved here: https://github.com/arrow-kt/arrow-core/issues/203

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
aballano: I believe one could achieve the same by just combining that possible Stream with the callback itself, no?

* `Stream.callback { ... }.merge(Stream.xxx)` 
* `Stream.callback { ... }.flatMap(Stream.xxx).parJoinUnbounded()`

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -0,0 +1,194 @@
...
+ */
+@UseExperimental(ExperimentalTypeInference::class)
+fun <A> Stream.Companion.cancellable(@BuilderInference f: suspend EmitterSyntax<A>.() -> CancelToken): Stream<A> =
+  force {
+    val q = Queue.unbounded<Chunk<A>>()
+    val error = UnsafePromise<Throwable>()
+    val cancel = Promise<CancelToken>()
+
+    Stream.bracketCase({
+      ForkAndForget { emitterCallback(f, cancel, error, q) }
```
aballano, line 124: @nomisRev is there a reason on why not using `ForkConnected` here instead?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
pablisco: That's ok if we have knowledge of the other stream outside. My thinking was if we require to do conditions inside the callback... 
So, depending on the value we get from a given callback we could start one or another.

One use case I can think of for the secuencial emit could be, if we have a lifecycle callback that generates a new stream of user events each time, ending the previous one. This way we could generate a seamless continuous stream of user events.

As for the parallel emit we could have something that listens to multiple UI elements (like selectable items) and would mean we can combine them all into a single stream.

Not a very good example that last one as it could be done as a merge...

Let's create a separate ticket and I'll investigate more useful use cases or drop it if we find it not that worth it :)

aballano:
I'll merge on green and solve any leftover questions/tickets separately

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/ops/CallbackTest.kt
pablisco: Moved here: https://github.com/arrow-kt/arrow-fx/issues/239

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
pablisco: Moved here: https://github.com/arrow-kt/arrow-fx/issues/240

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
nomisRev: The acquire step of `bracket` runs in an `uncancellable` manner, so there is no cancellation to hook into.
So here instead we acquire the `Fiber` as a resource, and we `cancel` it when our interruptable stream finishes.

nomisRev:
Thanks @pablisco for the tickets!!

PR #203, opened 2020-06-30
Fix timeOutOrNull to return on the  context

nomisRev:
- Fixes `timeOutOrNull` or null to return on the previous context.

@rachelcarmena @aballano
I suspect this was the reason for the flaky `Uncancellable back pressures timeoutOrNull` test.
So let's keep an eye on this together ;)


rachelcarmena:
Awesome, thanks @nomisRev :raised_hands: Just a small comment. I'm seeing a new `Arb.long` with a fix to calculate the range however the new tests still use `Arb.int`.

nomisRev:
@rachelcarmena the new `Arb.long` is for the currently flaky test [`Uncancellable back pressures timeoutOrNull`](https://github.com/arrow-kt/arrow-fx/blob/49e39d11fd1a78888d1505924395a68708955d86/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/BracketCaseTest.kt#L13), which uses `Arb.long` but is unchanged in this PR.

PR #214, opened 2020-07-02
Enviroment

nomisRev:
This PR proposes an [Enviroment] interface, that can be used instead of the bare bones runners which don't provide elegant syntax.

The Enviroment type offers the same runners as `UnsafeRun` currently offers for `F`.
It however also incorporates `CoroutineContext` and `asyncErrorHandler`.

- `asyncErrorHandler` would currently be called *only* for `cancellableF` when it's returned `CancelToken` throws an exception. [ref](https://github.com/arrow-kt/arrow-fx/blob/d12f4a062428fb3632bdae50be3fbd235008f9e1/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/builders.kt#L27).

- `CoroutineContext` defines an initial ctx for `unsafeRunSync` so we don't need an `EventLoop`. This however means that by default`unsafeRunSync` called from `main` runs on `ForkJoinPool`.
Alternative would be to start on `EmptyCoroutineContext` but then we'd return the user to `ForkJoinPool` after the first `sleep` or other parallel op.


settings.gradle
```diff
@@ -1,6 +1,7 @@
 rootProject.name = 'arrow-fx-lib'
 
 include 'arrow-fx'
+include 'arrow-docs'
```
rachelcarmena, line 4: :point_up_2: :wink: :pray:

nomisRev: 😅

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Runs and forgets the operation.
+   * This operation runs uncancellable.
+   */
+  fun <A> unsafeRunAsync(fa: suspend () -> A, e: (Throwable) -> Unit, a: (A) -> Unit): Unit
+
+  /**
+   * Runs the operation in a cancelable way
+   * Operation can be cancelled by invoking the returend [Disposable]
```
joramnv, line 57: Typo. 
Also maybe add dots to the end of these sentences like you did in the other comments.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -0,0 +1,85 @@
...
+ * Therefore it's advised to always run on [ComputationPool] which is the default setting.
+ *
+ * [Environment] also has an [asyncErrorHandler] which by default redirects to [Throwable.printStackTrace],
+ * no user flow errors will every be send here [CancelToken] exceptions might bubble up here when they cannot be redirect to the user.
+ *
+ * This [Environment] is meant to be used in Java frameworks, or frameworks that do not expose [suspend] edge-points.
+ *
+ * This contract could be elaborated on Android to provide automatic cancellation on Android LifecycleOwner.
+ */
+interface Environment {
```
joramnv, line 21: I really like the implementation. Can't wait to use it! :smiley:

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Environment that can run [suspend] programs using [startCoroutine] and [startCoroutineCancellable].
+ *
+ * An [Environment] runs on a certain [CoroutineContext] which is used to start the programs on.
+ * Since coroutines always return where they were started, this [CoroutineContext] also defines where you return after operators like [sleep] or [evalOn].
+ * Therefore it's advised to always run on [ComputationPool] which is the default setting.
+ *
+ * [Environment] also has an [asyncErrorHandler] which by default redirects to [Throwable.printStackTrace],
+ * no user flow errors will every be send here [CancelToken] exceptions might bubble up here when they cannot be redirect to the user.
```
joramnv, line 15: Typo.

nomisRev:
Thanks @joramnv for the code review! 🙌

arrow-docs/docs/fx/README.md
```diff
@@ -6,7 +6,8 @@ permalink: /fx/
 
 # Arrow Fx. Typed FP for the masses
 
-Arrow Fx is a next-generation Typed FP Effects Library that makes effectful and polymorphic programming first class in Kotlin, and acts as an extension to the Kotlin native suspend system.
+Arrow Fx is a next-generation Typed FP Effects Library that makes tracked effectful programming first class in Kotlin,
```
JorgeCastilloPrz, line 9: Perfect 👍

arrow-docs/docs/fx/README.md
```diff
@@ -87,158 +92,75 @@ suspend fun greet(): Unit {
...
-//sampleEnd
-fun main() {
-  println(greet())
-}
-```
+## Executing effectful programs
 
-When we capture suspended side effects as `IO` values with `effect`, we can pass them around and compose them until we are ready to apply the effects. At this point, nothing has happened because `greet` and `effect` are lazy values.
+The `greet` program is ready to run as soon as the user is ready to commit to an execution strategy that is either `blocking` or `non-blocking`.
+`blocking` execution strategies will block the current thread that's waiting for the program to yield a value, whereas `non-blocking` strategies will immediately return and perform the program's work without blocking the current thread.
```
JorgeCastilloPrz, line 98: This would make more sense to be part of the same paragraph I'd say.

arrow-docs/docs/fx/README.md
```diff
@@ -87,158 +92,75 @@ suspend fun greet(): Unit {
...
 
-#### Applying side effects with `!effect`
+Since both blocking and non-blocking execution scenarios perform side effects, we consider running effects as an `unsafe` operation.
```
JorgeCastilloPrz, line 100: Is `unsafe` here really code or not anymore given the new `Environment` api? if not I wouldn't use backticks.

arrow-docs/docs/fx/README.md
```diff
@@ -87,158 +92,75 @@ suspend fun greet(): Unit {
...
 
-We apply side effects with the operator `!`. Once you purify a side effect with `effect`, you can extract its value in a non-blocking way. `!effect` takes the suspended side effects `sayHello()` and `sayGoodbye()` and ensures the continuation context controls them before they get a chance to be executed. This ensures our effect compositions are pure and referentially transparent, and will only run at the edge.
+Arrow restricts the ability to run programs to extensions of the `Enviroment` type class.
```
JorgeCastilloPrz, line 102: I understand we'd likely suggest doing it, but does Arrow really **restrict** it? I can still run suspended programs without it can't I? Maybe *"Arrow promotes the ability..."* instead?

arrow-docs/docs/fx/README.md
```diff
@@ -87,158 +92,75 @@ suspend fun greet(): Unit {
...
 
-Note that running `greet()` in the previous example does not perform any effects because it returns a wrapped lazy value. Since invoking this function does not produce effects, we can be confident that `greet` is pure and referentially transparent, despite referring to the effects application.
+Usage of `unsafe` is reserved for the end of the world and may be the only impure execution of a well-typed functional program.
```
JorgeCastilloPrz, line 104: Is `unsafe` still correct here due to the new api? Maybe something like this or similar?

```suggestion
Usage of unsafe runner functions (like `unsafeRunSync` in this case) is reserved for the end of the world and may be the only impure execution of a well-typed functional program.
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -0,0 +1,86 @@
...
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.startCoroutine
+
+/**
+ * Environment that can run [suspend] programs using [startCoroutine] and [startCoroutineCancellable].
+ *
+ * An [Environment] runs on a certain [CoroutineContext] which is used to start the programs on.
+ * Since coroutines always return where they were started, this [CoroutineContext] also defines where you return after operators like [sleep] or [evalOn].
+ * Therefore it's advised to always run on [ComputationPool] which is the default setting.
```
JorgeCastilloPrz, line 12: Shouldn't we default to `EmptyCoroutineContext` as the kx coroutines builders like `async` or `launch`? Thinking on Android for example, you always wanna start stuff on the main thread then you decide which effects you want to offload to specific contexts like `ComputationPool` or `IOPool` with `evalOn` per occurrence, but you always want to get back to the caller thread which in Android that's the Main one. (For the majority of the cases). Also outside of Android, doesn't it make more sense to default to a no-op context like `EmptyCoroutineContext` so you stay on the caller thread for predictability?

nomisRev: No, we cannot default to `EmptyCoroutineContext` since that's an empty map and thus doesn't have a `ContinuationInterceptor`. As a result when you call an operator such as `evalOn` there is no way to intercept to return to the original context. In order for that to be possible we'd need launch an `EventLoop` on the current context, so we can later return there. Which of course also comes with a cost.

I.e. 

```
//main w/ EmptyCoroutineContext
evalOn(ComputationPool) {
  // forkjoinpool
} <-- // Can't go back to main since `EmptyCoroutineContext` cannot dispatch us back there
```

> you always wanna start stuff on the main thread

Typically on Android View setup needs to happen as fast as possible, and eagerly. Often this also has to happens in a different callbacks than where you call your business logic. I.e. `onCreateView`, `onViewCreated` & `onResume`.
So instead I prefer to launch my business logic on `ComputationPool`, and return to the `UI` if/when I need to using `evalOn(UI)`.

Alternatively if you can also launch on `Main` CoroutineContext for Android which would come at the price of a `post`. Unless you implement it with a check to see if it's already on `Main` then it would only `post` when returning from a different context like in the above example. We can however not default to `Android` Main here, but we could do so in an `AndroidEnviroment` in an Android module. Where can also provide nicer APIs for Android.

```
interface AndroidEnviroment : Enviroment {
  fun unsafeRunScoped(lifecycleOwner: LifecycleOwner, ...)
}
```

JorgeCastilloPrz: Got it. Regardless Android, I'm used to stay on caller thread until there's something computationally or IO expensive that requires jumping off to other pools in my program. Same for tests, you start execution on the test thread and then at some point and only if needed, you can jump off for a given effect (evalOn). Here, we do the other way around. We assume that anybody that runs a program by default (and uses `Environment`) wants the program to run in a `ComputationPool` by default and just jump off of this pool explicitly on some arbitrary places. I'm not 100% sure about that since feel like assuming a thread jump at the beginning for the user. But I believe we've had this conversation many times, so not a blocker for this PR at all 👍

nomisRev: This pattern is a pattern commonly on back-end. Maybe @raulraja can shed some more light.

We should also stay warry not to let any front-end concerns leak into Arrow Fx Coroutines. Therefore I intentionally left the implementation open for composition through an interface, the default implementation is a simple wrapper of the primitive runners.

Running your program on `Main` or similar single thread contexts comes with a hidden drawback that it's easy to keep it occupied with long computations without realizing it until you see freezing in the app, no?

aballano: > Running your program on `Main` or similar single thread contexts comes with a hidden drawback that it's easy to keep it occupied with long computations without realizing it until you see freezing in the app, no?

In general yes, on Android specifically there are logs and checks to prevent that by either crashing the app or by showing some cryptic logs... 

I agree with @nomisRev on this because by defaulting to a computation thread we eliminate the burden of the dev to remember to switch threads, because not doing so might or **might not** be noticeable during testing, but might come up on real and slower devices. So you only need to "remember" to post on UI, which you can very easily automate by creating some `evalOn(UI) { post(state) }` kind of functions.

arrow-docs/docs/fx/README.md
nomisRev: Yes, you're right it should be:

",we consider running effects with the `unsafeXX` operators unsafe", or I should remove the backticks. I think removing the backticks would make it more clear here, no?

arrow-docs/docs/fx/README.md
nomisRev: You're absolutely right. I was trying to emphasize that this is the only type that provides runners, and it should only be used at the edge.

arrow-docs/docs/fx/README.md
raulraja: No backticks makes It more clear

arrow-docs/docs/fx/README.md
JorgeCastilloPrz: Yep, for that I think "Arrow promotes..." is more accurate without losing any meaning.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
JorgeCastilloPrz: Sounds good, thanks for sharing thoughts folks 👍

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -0,0 +1,86 @@
...
+  fun asyncErrorHandler(e: Throwable): Unit
+
+  /**
+   * Execution strategy that will block the current thread that's waiting for the program to yield a value.
+   */
+  fun <A> unsafeRunSync(fa: suspend () -> A): A
+
+  /**
+   * Execution strategies that will immediately return and perform the program's work without blocking the current thread.
+   * This operation runs uncancellable.
```
pakoito, line 44: It may be worth adding a note about how exceptions are handled in this particular case.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -0,0 +1,86 @@
...
+
+  fun <A> unsafeRunAsyncCancellable(fa: suspend () -> A, e: (Throwable) -> Unit, a: (A) -> Unit): Disposable
+
+  companion object {
+    operator fun invoke(ctx: CoroutineContext = ComputationPool): Environment =
+      DefaultEnvironment(ctx)
+  }
+}
+
+internal class DefaultEnvironment(override val ctx: CoroutineContext) : Environment {
```
pakoito, line 71: I'd give it a `toString` identity so it can be identified in traces.

PR #222, opened 2020-07-06
Fix release after pure interruption

nomisRev:
When _pure_ interruption occurs, then `bracket` should be closed with `ExitCase.Cancelled`.

However when an error is thrown from the `release` function, then this should become the final result from the `Stream` since we do not want the `Stream` to silently finish when an error occurred.

This more the fix made in #210 into a separate PR.


arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/BracketTest.kt
```diff
@@ -234,6 +235,35 @@ class BracketTest : StreamSpec(spec = {
       }
     }
 
+    "propagate error from release" {
```
nomisRev, line 238: This was already passing but missing

PR #228, opened 2020-07-13
Remove Try from tests

aballano:
1 of 2 to unlock https://github.com/arrow-kt/arrow-core/pull/149


arrow-fx-rx2/src/test/kotlin/arrow/fx/FlowableKTests.kt
```diff
@@ -153,12 +143,12 @@ class FlowableKTests : RxJavaSpec() {
 
     "FlowableK cancellable should cancel CancelToken on dispose" {
       Promise.uncancellable<ForFlowableK, Unit>(FlowableK.async()).flatMap { latch ->
-          FlowableK {
-            FlowableK.cancellable<Unit>(fa = {
-              latch.complete(Unit)
-            }).flowable.subscribe().dispose()
-          }.flatMap { latch.get() }
-        }.value()
+        FlowableK {
```
nomisRev, line 146: Was this not properly checked by `ktlintCheck`? 🤔

arrow-fx-test/src/main/kotlin/arrow/fx/test/eq/EqK.kt
```diff
@@ -63,3 +64,23 @@ fun Schedule.Decision.Companion.eqK(): EqK<DecisionPartialOf<Any?>> = object : E
...
+  val bb = try {
+    fb()
+  } catch (err: Throwable) {
+    err
+  }
+  return when {
+    aa is TimeoutException -> throw aa
+    bb is TimeoutException -> throw bb
+    aa is Exception && bb is Exception -> throwableEq().run { aa.eqv(bb) }
+    aa is Exception || bb is Exception -> false
```
nomisRev, line 83: Should we re-throw here?

If we don't bubble up the exception, wouldn't we otherwise get an `expected true but found false` assertion here from `forAll`?

aballano: In theory this should be equivalent to https://github.com/arrow-kt/arrow-fx/pull/228/files#diff-4399f003c25a9b249e3fab432f659bd4L191

arrow-fx-rx2/src/test/kotlin/arrow/fx/FlowableKTests.kt
aballano: Seems not, no idea why tho 😕

PR #231, opened 2020-07-17
Bump Kotest

aballano:
Remove explicit test lib declaration
Change kotlin to jdk8


arrow-fx/src/test/kotlin/arrow/fx/ArrowFxSpec.kt
```diff
@@ -1,9 +1,9 @@
 package arrow.fx
 
 import arrow.core.test.UnitSpec
-import io.kotlintest.properties.Gen
-import io.kotlintest.properties.PropertyContext
-import io.kotlintest.properties.forAll
+//import io.kotest.property.Arb
+//import io.kotest.property.PropertyContext
+//import io.kotest.property.forAll
```
rachelcarmena, line 6:
```suggestion
// import io.kotest.property.Arb
// import io.kotest.property.PropertyContext
// import io.kotest.property.forAll
```

because of Ktlint failures

PR #232, opened 2020-07-17
Fix constant cancellation & cancellation race condition acquiring a resource

nomisRev:
This PR adds a test suite for cancellation.

It exposed two bugs:
 - That cancellation is not properly checked in interruption check points.
 - That there can occur a cancellation race condition can occur between acquiring a resource, and atomatically registering it into the scope.


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/Scope.kt
```diff
@@ -174,19 +166,16 @@ class Scope private constructor(
...
-            Either.Left(eith.swap().getOrElse { AcquireAfterScopeClosed })
-        }
-
-      is Either.Left -> eith
+    val conn = coroutineContext.connection()
+    val scope = ScopedResource()
+    return Either.catch(fr).flatMap { resource ->
+      scope.acquired { ex: ExitCase -> release(resource, ex) }.map { registered ->
+        state.modify {
+          if (conn.isCancelled() && registered) Pair(it, suspend { release(resource, ExitCase.Cancelled) })
```
nomisRev, line 174: We need to check for cancellation while registering the resource into the scope, if while registering the resource a cancellation signal occurs than we need to release the resource immediately with `ExitCase.Cancelled`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ScopedResource.kt
```diff
@@ -56,23 +60,37 @@ internal class ScopedResource {
...
-      } else {
-        val attemptFinalizer: suspend (ExitCase) -> Either<Throwable, Unit> =
-          { ec -> Either.catch { finalizer(ec) } }
-        // either state is open, or leases are present, either release or `Lease#cancel` will run the finalizer
-        Pair(s.copy(finalizer = attemptFinalizer), suspend { Either.Right(true) })
+  suspend fun acquired(finalizer: suspend (ExitCase) -> Unit): Either<Throwable, Boolean> {
+    val conn = coroutineContext.connection()
+    return state.modify { s ->
+      when {
+        conn.isCancelled() -> {
```
nomisRev, line 67: We need to check for cancellation while registering the resource into the scope, if while registering the resource a cancellation signal occurs than we need to release the resource immediately with `ExitCase.Cancelled`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/Compiler.kt
```diff
@@ -118,7 +119,7 @@ internal fun <O> interruptBoundary(
   }
 
 internal suspend inline fun interruptGuard(scope: Scope): Result<Any?>? =
-  when (val isInterrupted = scope.isInterrupted()) {
+  when (val isInterrupted = scope.isInterrupted().also { cancelBoundary() }) {
```
nomisRev, line 122: Check for cancellation in `interruptGuard`

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ScopedResource.kt
aballano: Since you're not gonna use the connection anywhere else, wouldn't be better to inline it here? this way you let less space for racing between variable declaration and few functions called in between

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/Scope.kt
aballano: Same here, conn could be inlined

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/predef-test.kt
```diff
@@ -81,6 +80,46 @@ suspend fun assertCancellable(f: suspend () -> Unit): Unit {
...
+    )
+  }
+
+  start.get()
+  fiber.cancel()
+  p.get() shouldBe ExitCase.Cancelled
+}
+
+@JvmName("assertStreamCancellable")
+suspend fun <A> assertCancellable(fa: (latch: Promise<Unit>) -> Stream<A>): Unit {
```
aballano, line 105: If we would have this functions here as utility functions I think they should be generic enough to use them in more places and exclude the assertions from here, otherwise we're hiding the actual testing from the test classes.

Alternatively you could move them to the CancellationTest

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ScopedResource.kt
nomisRev: `isCancelled` can actually be called multiple times here, that's why the connection should be cached.

All operations on `atomic` and `Atomic` that take a function are prone to be called multiple times until it was able to succesfully update using `compareAndSet` since `set(f(get()))` is not an atomic safe operation.

That's also why `Atomic` (and `Ref`) don't take `suspend` (or `F`) signatures. It's advised against to call effectful code inside atomic updates, since slow updates would have a very low chance of a succesful atomic update.
Therefore we also use `modify` here and use a `suspend` lambda to defer the operation to happen outside the atomic update.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/predef-test.kt
nomisRev: How would you make it more generic? You can still add your own assertions arround, as done [here](https://github.com/arrow-kt/arrow-fx/pull/232/files#diff-19c8e5cf480f27c6f5ec7dc44549de24R36). 

We can definitely move them to `CancellationTest`. We haven't discussed how we want to organise the test suite, now it's still mixed between a bunch of stuff in `StreamTest`, `ParJoinTest`, `BracketTest`, etc

I personally like grouping tests per combinator (or the `Concurrent` onces) so it can serve as a specification of it's behavior, like a law test suite.
If we'd do that I think we should add the `cancellation`/`interruption` tests there as well for it to be a full spec of it's behavior.

aballano: @nomisRev ?

PR #244, opened 2020-08-05
Add iterative terminal operation and a shortcut

pablisco:
Very small PR to add a new terminal operation that "iterates" over each value emitted.

I've also added a shortcut in case we want to avoid calling compile every time. 

Shall I add the shortcuts for the other terminal operations?

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
@@ -14,23 +14,7 @@ package arrow.fx.coroutines.stream
     compiler(mutableSetOf()) { acc, ch -> acc.apply { addAll(ch.toList()) } }
 
   suspend fun drain(): Unit =
-    foldChunks(Unit) { _, _ -> Unit }
-
-  /**
-   * Similar to [drain] but applying an effect on each of the objects emitted.
-   * 
-   * This equivalent to doing:
-   * 
-   * ```
-   * stream.effectTap { **operation** }
-   *   .compile()
-   *   .drain()
-   * ```
-   * 
-   * The main difference, apart from being shorter, is that the operation is applied after we compile the stream into [TerminalOps] 
-   */
-  suspend fun forEach(block: (O) -> Unit): Unit =
-    foldChunks(Unit) { _, ch -> ch.forEach(block) }
+    compiler(Unit) { _, _ -> Unit }
 
   /**
    * Compiles this stream in to a value,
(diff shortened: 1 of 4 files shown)


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOpsShortcuts.kt
```diff
@@ -0,0 +1,7 @@
+package arrow.fx.coroutines.stream
+
+/**
+ * This is a shortcut for [compile] + [TerminalOps.forEach]
+ */
+suspend fun <O> Stream<O>.forEach(block: (O) -> Unit): Unit =
```
nomisRev, line 6: The design of `TerminalOps` was made to make all terminal operators very explicit, since this is where you exit the streaming world. Which you should typically never do this in functional streaming applications since you want to use `Stream` all the way to the edge in those cases.

Adding this would nullify the whole design of `TerminalOps`, and `TerminalResourceOps`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -16,6 +16,22 @@ package arrow.fx.coroutines.stream
...
+   * 
+   * ```
+   * stream.effectTap { **operation** }
+   *   .compile()
+   *   .drain()
+   * ```
+   * 
+   * The main difference, apart from being shorter, is that the operation is applied after we compile the stream into [TerminalOps] 
+   */
+  suspend fun forEach(block: (O) -> Unit): Unit =
```
nomisRev, line 32: 😱 `forEach`

I think we should implement `fold` here as an overload of `foldChunks` instead of `forEach`, which is in line with the APIs and naming we use in Arrow Core.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOpsShortcuts.kt
pablisco: Good point, I'll remove the shortcut 😁

pablisco:
I've replaced the forEach for a fold, I guess it's more explicit to do `fold(Unit) { _, it -> do(it) }`

pablisco:
@nomisRev This is probably no longer relevant after https://github.com/arrow-kt/arrow-fx/pull/317 right?

nomisRev:
Hey @pablisco, no this will no longer be relevant if #317 goes through.

pablisco:
Following up here: https://github.com/arrow-kt/arrow-fx/issues/325

PR #246, opened 2020-08-07
Outstanding dequeues not completed on tryOffer1

nomisRev:
The `update` function in `PubSub` was not correctly invoking the completion for outstanding dequeues. 

To fix this I replaced the usages of `Promise` to `UnsafePromise` so we can complete the promises in a non-suspending manner. This allows for both `modify` and `update` which can modify the state with `suspend`, and `non-suspend` support.

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/PubSub.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/PubSub.kt
@@ -5,10 +5,9 @@ import arrow.core.None
 import arrow.core.Option
 import arrow.core.Some
 import arrow.fx.coroutines.ExitCase
-import arrow.fx.coroutines.ForkConnected
 import arrow.fx.coroutines.IQueue
-import arrow.fx.coroutines.Promise
 import arrow.fx.coroutines.Token
+import arrow.fx.coroutines.UnsafePromise
 import arrow.fx.coroutines.guaranteeCase
 import arrow.fx.coroutines.stream.Chunk
 import arrow.fx.coroutines.stream.Stream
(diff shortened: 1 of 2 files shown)


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/PubSub.kt
```diff
@@ -79,15 +78,15 @@ internal interface Subscribe<A, Selector> {
 
 internal interface PubSub<I, O, Selector> : Publish<I>, Subscribe<O, Selector> {
 
-  data class Publisher<A>(val token: Token, val i: A, val signal: Promise<Unit>) {
-    suspend fun complete(): Unit {
-      ForkConnected { signal.complete(Unit) }
```
nomisRev, line 84: It's not necessary to launch completion in a `Fiber` since the `join` operations already `intercept` before returning, meaning they always get properly re-scheduled before returning from `join`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/PubSub.kt
```diff
@@ -566,9 +565,9 @@ internal class DefaultPubSub<I, O, QS, S>(private val strategy: PubSub.Strategy<
   fun <X> update(f: (PubSub.PubSubState<I, O, QS, S>) -> Pair<PubSub.PubSubState<I, O, QS, S>, X>): X =
     state.modify { ps ->
       val (ps1, result) = f(ps)
-      val (ps2, _) = loop(ps1) { Unit }
-      Pair(ps2, result)
-    }
+      val (ps2, action) = loop(ps1) { Unit }
+      Pair(ps2, { action.invoke(); result })
```
nomisRev, line 569: This was the bug fix, in case an outstanding offer was found. Its completion needs to be invoked.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/PubSub.kt
```diff
@@ -659,13 +658,13 @@ internal class DefaultPubSub<I, O, QS, S>(private val strategy: PubSub.Strategy<
    */
   private tailrec fun loop(
     ps: PubSub.PubSubState<I, O, QS, S>,
-    action: suspend () -> Unit
-  ): Pair<PubSub.PubSubState<I, O, QS, S>, (suspend () -> Unit)> {
```
nomisRev, line 663: The actions running in the `loop` no longer need to be suspending since we rely on `UnsafePromise` for signaling.

Instead the functionality of `suspend` is exposed through the `modify` operator.

PR #247, opened 2020-08-10
Doc: show benchmarks report

rachelcarmena:
And then it would be necessary to add this option to the sidebar:

```diff
+  - title: Reports
+
+    nested_options:
+
+      - title: JMH
+        url: /reports/jmh/report.html
```


raulraja:
Closing this as we are actually gonna remove the benchmarks module since our ops are unrelated to those found in Scala and others. We are already certain we are faster in the scenarios we care since other libs care about indirect flatMap related composition and we don't given we are now completely based on suspend functions as we remove IO.

PR #248, opened 2020-08-11
Improve Arrow Fx Coroutines docs for release

nomisRev:
Another iteration over the docs.
This PR goes together with https://github.com/arrow-kt/arrow-site/pull/64 which updates the sidebar of the Arrow Fx website to promote Arrow Fx Coroutines for the next version.

### Quickstart section

Updated the second page from the Arrow Fx website Quickstart to Arrow Fx Coroutines, this leaves the 3rd page which talks about tagless effects.

This could become a page that in more detail explains the integrations and cancellation in the standard library, etc.
A fourth page could be added to the quickstart which covers the `IO` vs `suspend` section currently in the [README.MD of Arrow Fx Coroutines](https://github.com/arrow-kt/arrow-fx/tree/master/arrow-fx-coroutines#ioa-vs-suspend----a). It could also include a small migration guide from `IO` to `suspend` as [today was asked on Slack](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1597146639290000)

### Data type section (non-blockers for PR)

- [ ] CircuitBreaker only has API documentation. It was inspired by Monix's CircuitBreaker which in turn was inspired by Akka's CircuitBreaker. Monix's seems to overlap with Akka's docs, so perhaps with the correct credits we can do the same. [link](https://monix.io/docs/current/catnap/circuit-breaker.html).

- [ ] API docs are a very low profile and have little to no styling. I'm not sure how or if we need to improve our API docs, or if we should promote certain functions in a different manner. [parMapN](https://arrow-kt.io/docs/next/apidocs/arrow-fx-coroutines/arrow.fx.coroutines/par-map-n.html), [parTupledN](https://arrow-kt.io/docs/next/apidocs/arrow-fx-coroutines/arrow.fx.coroutines/par-tupled-n.html), [raceN](https://arrow-kt.io/docs/next/apidocs/arrow-fx-coroutines/arrow.fx.coroutines/race-n.html).

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
@@ -78,10 +78,13 @@ suspend fun <A> onCancel(
 }
 
 /**
- * Executes the given [finalizer] when the source [fa] is finished, either in success or in error, or if cancelled.
+ * Guarantees execution of a given [finalizer] after [fa] regardless of success, error or cancellation.
  *
  * As best practice, it's not a good idea to release resources via [guarantee].
- * Prefer [Resource] or [bracket] for the acquisition and release of resources.
+ * since [guarantee] doesn't properly model acquiring, using and releasing resources.
+ * It only models scheduling of a finalizer after a given suspending program,
+ * so you should prefer [Resource] or [bracket] which captures acquiring,
+ * using and releasing into 3 separate steps to ensure resource safety.
  *
  * @param fa program that you want to register handler on
  * @param finalizer handler to run after [fa].
@@ -93,11 +96,14 @@ suspend fun <A> guarantee(
 ): A = guaranteeCase(fa) { finalizer.invoke() }
 
 /**
- * Executes the given [finalizer] when the source [fa] is finished, either in success or in error, or if cancelled, allowing
+ * Guarantees execution of a given [finalizer] after [fa] regardless of success, error or cancellation., allowing
  * for differentiating between exit conditions with to the [ExitCase] argument of the finalizer.
  *
  * As best practice, it's not a good idea to release resources via [guaranteeCase].
- * Prefer [Resource] or [bracketCase] for the acquisition and release of resources.
+ * since [guaranteeCase] doesn't properly model acquiring, using and releasing resources.
+ * It only models scheduling of a finalizer after a given suspending program,
+ * so you should prefer [Resource] or [bracketCase] which captures acquiring,
+ * using and releasing into 3 separate steps to ensure resource safety.
  *
  * @param fa program that you want to register handler on
  * @param finalizer handler to run after [fa].
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
@@ -56,6 +56,17 @@ suspend fun <A> uncancellable(f: suspend () -> A): A =
     }
   }
 
+/**
+ * Registers an [onCancel] handler after [fa].
+ * [onCancel] is guaranteed to be called in case of cancellation, otherwise it's ignored.
+ *
+ * Useful for wiring cancellation tokens between fibers, building inter-op with other effect systems or testing.
+ *
+ * @param fa program that you want to register handler on
+ * @param onCancel handler to run when [fa] gets cancelled.
+ * @see guarantee for registering a handler that is guaranteed to always run.
+ * @see guaranteeCase for registering a handler that executes for any [ExitCase].
+ */
 suspend fun <A> onCancel(
   fa: suspend () -> A,
   onCancel: suspend () -> Unit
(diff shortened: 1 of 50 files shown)


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -66,22 +77,143 @@ suspend fun <A> onCancel(
   }
 }
 
+/**
+ * Executes the given [finalizer] when the source [fa] is finished, either in success or in error, or if cancelled.
+ *
+ * As best practice, it's not a good idea to release resources via [guarantee].
```
raulraja, line 83: we should probably mention here why it's not a good idea.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -66,22 +77,143 @@ suspend fun <A> onCancel(
...
 suspend fun <A> guaranteeCase(
   fa: suspend () -> A,
-  release: suspend (ExitCase) -> Unit
-): A = bracketCase({ Unit }, { fa.invoke() }, { _, ex -> release(ex) })
+  finalizer: suspend (ExitCase) -> Unit
+): A = bracketCase({ Unit }, { fa.invoke() }, { _, ex -> finalizer(ex) })
 
+/**
+ * Meant for specifying tasks with safe resource acquisition and release in the face of errors and interruption.
+ * It would be the the equivalent of `try/catch/finally` statements in mainstream imperative languages for resource
```
raulraja, line 113:
```suggestion
 * It would be the equivalent of an async capable `try/catch/finally` statements in mainstream imperative languages for resource
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -66,22 +77,143 @@ suspend fun <A> onCancel(
...
+ *   1. acquisition
+ *   2. consumption
+ *   3. releasing
+ *
+ * 1. Resource acquisition is **NON CANCELLABLE**.
+ *   If resource acquisition fails, meaning no resource was actually successfully acquired then we short-circuit the effect.
+ *   Reason being, we cannot [release] what we did not `acquire` first. Same reason we cannot call [use].
+ *   If it is successful we pass the result to stage 2 [use].
+ *
+ * 2. Resource consumption is like any other `suspend` effect. The key difference here is that it's wired in such a way that
```
raulraja, line 170: Excellent docs and clarifications!

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/cancelBoundary.kt
```diff
@@ -6,6 +6,28 @@ import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
...
+ * //sampleStart
+ * suspend fun forever(): Unit {
+ *   while(true) {
+ *     println("I am getting dizzy...")
+ *     cancelBoundary() // cancellable computation loop
+ *   }
+ * }
+ *
+ * suspend fun main(): Unit {
+ *   val fiber = ForkConnected {
```
raulraja, line 22: I'm not clear about the name ForkConnected. Could you elaborate a bit about why the Connected part?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
nomisRev: Should we repeat this in `bracket`, I am unsure how we can improve the documentation of these top-level functions.

> API docs are a very low profile and have little to no styling. I'm not sure how or if we need to improve our API docs, or if we should promote certain functions in a different manner. parMapN, parTupledN, raceN.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/cancelBoundary.kt
nomisRev: I added docs for `ForkConnected` with an example after this comment :D [link](https://github.com/arrow-kt/arrow-fx/pull/248/files#diff-5b3bc4412001765ff242219bda005d13R41)

##### Reasoning for new `ForkConnected` and `ForkScoped` functions 

In Arrow Fx IO we only had `fork` which launches a `Fiber` without any cancellation connection control. Meaning you *always* leak unless you launch a `Fiber` inside `Resource` or `bracketCase` as the `acquire` step.
This makes `fork` quite complicated and easily results in cancellation bugs etc. Those semantics of IO's `fork` is now an combinator called `ForkAndForget.

Some more info [here](https://github.com/arrow-kt/arrow-fx/tree/master/arrow-fx-coroutines#racepairracetripleforkandforget). We should probably also document these, but this comes back to trying to figure out a way to add higher-level documentation to top-level functions besides function docs. Or do you think we should include all of that there?

nomisRev:
@raulraja Thanks for the review! Did I address all your comments? Do you have any feedback on my two questions (checkboxes)?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/dispatchers.kt
```diff
@@ -9,35 +9,42 @@ import kotlin.coroutines.Continuation
...
+ * This pool is prone to cause [OutOfMemoryError] since the pool size is unbounded.
+ */
+val IOPool: CoroutineContext =
+  Executors.newCachedThreadPool { r ->
+    Thread(r).apply {
+      name = "io-arrow-kt-worker-${IOCounter.getAndIncrement()}"
+      isDaemon = true
+    }
+  }.asCoroutineContext()
+
```
rachelcarmena, line 39:
```suggestion

```
Awesome @nomisRev !! I think KLint is complaining about this extra line!

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/cancelBoundary.kt
```diff
@@ -6,6 +6,28 @@ import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
 /**
  * Checks for cancellation,
  * when cancellation occurs this coroutine will suspend indefinitely and never continue.
```
PhBastiani, line 8: The example is clear ... but, the textual explanation does not seem sufficient to understand the purpose of cancelBoundary

rachelcarmena:
About:

>     * API docs are a very low profile and have little to no styling. I'm not sure how or if we need to improve our API docs, or if we should promote certain functions in a different manner. 

Let's try the new Dokka which introduces a modern HTML format!!

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/cancelBoundary.kt
nomisRev: What about the following instead?  I found this terribly hard to document.

"Inserts a cancellable boundary — it checks for the cancellation status of the suspend-loop and does not allow the continuation to keep executing in the case cancellation happened."

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -66,22 +77,149 @@ suspend fun <A> onCancel(
...
+ *
+ * @param use is the action to consume the resource and produce a result.
+ * Once the resulting suspend program terminates, either successfully, error or disposed,
+ * the [release] function will run to clean up the resources.
+ *
+ * @param release is the action that's supposed to release the allocated resource after `use` is done, irregardless
+ * of its exit condition.
+ *
+ * ```kotlin:ank:playground
+ * import arrow.fx.IO
```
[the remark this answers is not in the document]
nomisRev: How did this ever pass Ank??? 😱

nomisRev:
@PhBastiani Thanks for the amazing review and taking the time to go through the PR! 👍 
Please let me know if you think I address all your comments.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/cancelBoundary.kt
PhBastiani: Maybe:

In a cancelable environment, we need to add mechanisms to react when a cancellation is triggered : a cancel boundary inserts checking point in suspendable methods. At runtime, a cancel boundary checks for the cancellation status of the suspend-loop and does not allow the continuation to keep executing in the case a cancellation was triggered.

Or, something like that

nomisRev: Thank you @PhBastiani! That is a much better improvement over what I did 😅

PR #249, opened 2020-08-11
Revert parTupledN and parMapN derivation

nomisRev:
Currently `parMapN` is derived from `parTupledN` meaning that for `parMapN` you always do a redundant allocation, this was originally due to how we derived `Applicative` functions. So this also reduces `parTraverse` with `n` allocation for a list of size `n`.

Additionally, we always deal with tupled functions in `parMapN` that forces users to always destructure. Additionally, it disallows us to use method references in a convenient way. [reference](
https://github.com/arrow-kt/arrow-fx/pull/248/files?file-filters%5B%5D=.kt&file-filters%5B%5D=.md#diff-c62032ca8b6b0243e4ffb313cbc8df73R69)


aballano (approved):
👌

PR #250, opened 2020-08-12
Queue: support same behavior as Arrow Fx's IO Queue

nomisRev:
This PR renames `Queue.circularBuffer` to `Queue.sliding` and align behavior with `Queue.sliding` as we have today in Arrow Fx. As discussed with @danimontoya based on his usages.
This also algins the behavior with `Channel.conflated` from KotlinX.

Also adds missing support for dropping Queues.

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/Queue.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/Queue.kt
@@ -150,6 +150,7 @@ interface Queue<A> : Enqueue<A>, Dequeue1<A>, Dequeue<A> {
     fun <A> unsafeSliding(maxSize: Int): Queue<A> =
       fromStrategy(Strategy.sliding(maxSize))
 
+    /** Creates a queue which stores the first `maxSize` enqueued elements and which never blocks on enqueue. */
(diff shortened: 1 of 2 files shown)


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/concurrent/Queue.kt
```diff
@@ -144,11 +144,17 @@ interface Queue<A> : Enqueue<A>, Dequeue1<A>, Dequeue<A> {
...
-      fromStrategy(Strategy.circularBuffer(maxSize))
+    suspend fun <A> sliding(maxSize: Int): Queue<A> =
+      fromStrategy(Strategy.sliding(maxSize))
 
-    fun <A> unsafeCircularBuffer(maxSize: Int): Queue<A> =
-      fromStrategy(Strategy.circularBuffer(maxSize))
+    fun <A> unsafeSliding(maxSize: Int): Queue<A> =
+      fromStrategy(Strategy.sliding(maxSize))
+
+    suspend fun <A> dropping(maxSize: Int): Queue<A> =
```
aballano, line 153: Could you include a small kdoc here as well? 🙏

nomisRev: Of course! Thanks!

PR #251, opened 2020-08-12
Update README.MD

PhBastiani:
Additionnal notes : Atomic & ConcurrenVar

### DIFF
--- a/arrow-fx-coroutines/README.MD
+++ b/arrow-fx-coroutines/README.MD
@@ -356,6 +356,7 @@ This code snippet isn’t very useful, except to show how `Ref/Atomic` can be us
 (i.e. a `ConcurrentVar` is a mutable concurrent safe variable which is either `empty` or contains a typed `single value`).
 
 ```kotlin:ank
+tailrec
 fun sum(state: MVar<ForIO, Int>, list: List<Int>) : IO<Int> =
     IO.fx {
         when {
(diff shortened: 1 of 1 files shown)


nomisRev:
Thanks @PhBastiani!

This is great 👏 👏 No work on this has been done yet, but I proposed something in #248 which is preparing the docs for release. I'd love your feedback there as well!

> This could become a page that in more detail explains the integrations and cancellation in the standard library, etc.
A fourth page could be added to the quickstart which covers the IO vs suspend section currently in the README.MD of Arrow Fx Coroutines. It could also include a small migration guide from IO to suspend as today was asked on Slack

arrow-fx-coroutines/README.MD
```diff
@@ -315,13 +315,82 @@ suspend main(): Unit {
...
+                state.put(cur + list[0]).flatMap { _ -> sum(state, list.tail()) }
+            }.bind()
+        }
+    }
+```
+
+becomes
+ 
+```kotlin:ank
+suspend fun sum(state: ConcurrentVar<Int>, list: List<Int>) {
```
nomisRev, line 372: This function can be `tailrec` and then it will be stacksafe ;)

PhBastiani: Yes thx

rachelcarmena:
Global checks were cancelled because they aren't necessary. Thanks @PhBastiani !!

PR #260, opened 2020-08-18
Remove TerminalOps from Stream

### LINKED ISSUE #245: ["Discussion"] Stream TerminalOps

`0.11.0-SNAPSHOT` 

What would you like to see?

Less redundant solution than `TerminalOps`, we now duplicate some operations in `TerminalOps` and in the `Stream` API.
Both offer the same behavior, with slightly different signatures. See #244 

I.e. `s.foldChunk`/`s.compile().foldChunks`, `s.fold`/`s.compile().fold` , etc

Discussed proposal with @pablisco on Slack.

Currently some operations such as `fold`, `foldChunks`, etc are defined both on `Stream<A>` and `TerminalOps<A>`.

```kotlin
fun <A, B> Stream<A>.fold(b: B, f: (A, B) -> B): Stream<B>
suspend fun <A, B> TerminalOps<A>.fold(b: B, f: (A, B) -> B): B
```

The first signatures is not properly typed since it doesn't return a `Stream<B>` but a single value `B` wrapped in a `Stream`, so exposing as `suspend fun <A, B> Stream<A>.fold(b: B, f: (A, B) -> B): B` would be more accurate.

This would allow us to remove `TerminalOps`, `TerminalResourceOps` and remove duplicated operators between the 3.

nomisRev:
Removes `TerminalOps` in favor of more precises signatures.

Closes #245.

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ParJoin.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/ParJoin.kt
@@ -163,7 +163,6 @@ internal suspend fun signalResult(done: SignallingAtomic<Option<Option<Throwable
  *       }
  *     }
  *     .parJoin(10, IOPool)
- *     .compile()
  *     .drain()
  * //sampleEnd
  * ```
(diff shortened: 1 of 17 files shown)


danimontoya (approved):
Less verbose and closer to Flow API! Looking good @nomisRev thank you!! 🎉

aballano (approved):
neat 👍

PR #262, opened 2020-08-21
Duration Comparison

red-avtovo:
Now you can compare durations like `a<b` or `b>a`


red-avtovo:
Is it ready to be merged?

rachelcarmena:
> Is it ready to be merged?

Yes, thanks @red-avtovo !! Sorry, let's update it with `master` branch again and it will be merged afterwards.

rachelcarmena:
> We should at some point move this to an `Order` instance for `Duration` so that we have all the order laws tested (the test here for comparison would be included there as well).
> 
> `Order` can be implemented with either `compare` or `compareTo` so this would only add the extra work of adding an instance^^
> 
> Also the fx-coroutines api could use this as well right?

Please @1Jajen1 , create an issue with this proposal in order not to miss it. Thanks!! :raised_hands:

PR #269, opened 2020-08-25
Add some docs for Queue and tests

nomisRev:
Some improvements based on the work today with @danimontoya on grpc Kotlin.

### DIFF
--- a/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/concurrent/QueueTest.kt
+++ b/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/concurrent/QueueTest.kt
@@ -2,9 +2,12 @@ package arrow.fx.coroutines.stream.concurrent
 
 import arrow.core.Option
 import arrow.fx.coroutines.ForkAndForget
+import arrow.fx.coroutines.ForkConnected
 import arrow.fx.coroutines.Promise
 import arrow.fx.coroutines.StreamSpec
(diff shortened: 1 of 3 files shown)


arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/concurrent/QueueTest.kt
```diff
@@ -188,4 +192,94 @@ class QueueTest : StreamSpec(spec = {
...
+        .concurrently(
+          s.noneTerminate()
+            .through(q.enqueue())
+        ).toList() shouldBe s.toList()
+    }
+  }
+
+  "Queue.bounded(0) with outstanding taker cannot offer" {
+    val q = Queue.bounded<Int>(0)
+    ForkConnected { q.dequeue1() }
```
danimontoya, line 256: Is this needed?

danimontoya (approved):
Thank you @nomisRev !!! 👍

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/concurrent/QueueTest.kt
nomisRev: That is to test that even tho someone is trying to take from the `Queue`, that `offer` still fails. We can't assume from the test that it won't do that, right? 😅

aballano (approved):
👌

PR #270, opened 2020-08-26
STM implementation

1Jajen1:
This pr implements software-transactional-memory in kotlin.

This allows writing transactions:
```kotlin
val counter = TVar.new(0)
val transaction = stm {
  if (counter.read() > 100) counter.write(0) else counter.write(counter.read() + 1)
}
// running
val res = atomically(transaction)
```
No matter how many threads run this transaction the value will always increase consistently. The above example will only ever do one read and it will never break the invariant of always having values between 0 and 101.

Similarly this also never breaks the invariants and is race-condition free:
```kotlin
suspend fun STM.moveTo(t1: TVar<Int>, t2: TVar<Int>, amount: Int): Unit {
  val a1 = t1.read()
  val a2 = t2.read()
  if (a1 - amount < 0) retry()
  else {
    t1.write(a1 - amount)
    t2.write(a2 + amount)
  }
}
// this is a bit shorter using check instead^^
suspend fun STM.moveTo2(t1: TVar<Int>, t2: TVar<Int>, amount: Int): Unit {
  val a1 = t1.read()
  val a2 = t2.read()
  check(a1 - amount >= 0)
  t1.write(a1 - amount)
  t2.write(a2 + amount)
} 
```
Here we use the primitive `retry` to rerun the transaction on the spot. This will actually suspend until either t1 or t2 changes.

There is also another primitive `orElse` which can be used to attach a fallback should a transaction fail.
```kotlin
val t = stm { retry() } orElse { 10 }
atomically(t) // 10
```

Now to the fun bits:
- This does not offer `unsafeSuspendToSTM` like GHC does with `unsafeIOToSTM`, not sure if we should even do that... It would certainly complicate things a bit.
- The locking bits are handwritten and thus all prone to be terribly bad, please check those implementations :see_no_evil:
- Locks are only taken for write commits. Transactions never wait for locks during commit, only on reads, which means they never block while holding locks themselves and thus there are no deadlocks. Also commits abort and retry when they try to lock an already locked var.
- I have no idea on how this works with cancellation, it would be great if committing a transaction would be uncancelable

Missing for a merge:
- [x] Documentation for STM itself 
- [x] Documentation for the datatypes and their methods
- [ ] ~Benchmarks (This is a maybe but it wouldn't hurt)~ Not happening in this pr. Although I had some benchmarks which show good results I felt like those need some more work to provide meaningful results


1Jajen1:
Apart from docs and exception handling inside transactions this should now be ready for review :tada: 

@nomisRev can you check whether or not cancellation might be problematic when committing? (a transaction never yields mid commit so I don't know if that is even possible)

Also at some point, either when this is merged or after, there should be benchmarks to see how this performs compared to alternatives, but I'll need some help setting that up (especially since the only good implementations to compare to are both written in scala and I haven't written a single line of scala code yet :sweat_smile:)

nomisRev:
I'm wondering if we should move such implementations to a different module than the `core` module Arrow Fx Coroutines. I'm thinking we should actually do the same with the FS2 port.

1Jajen1:
> I'm wondering if we should move such implementations to a different module than the `core` module Arrow Fx Coroutines. I'm thinking we should actually do the same with the FS2 port.

That sounds like a good idea :+1: As long as that is easy to do with how our docs are set up I don't see a good reason to not split this and the streams lib off.

1Jajen1:
This pr is now basically done. I need some documentation for TMap and TSet but that is easy to do.

I have some benchmarks set up and they show consistently faster speeds than zio stm, but they need some work to provide better value. (Writing good benchmarks that actually have meaning is hard ^^)
I could also use some help from someone with a scala background when it comes to adding benchmarks for other scala libs and reviewing the current ones. (I'll probably leave them out of this pr and do them in another one because this one is quite large already).

In any case this could use a review, not necessarily of everything, just reviews for docs would already be quite nice as there are likely things to improve :)

This pr now has a few primary parts:
- The STM interface which has most of the public api of stm datatypes and includes almost all documentation
- The STMTransaction and TVar implementations which are the two building blocks of STM and tests
- All the stm datatypes (except TVar). Those are all built upon the STM primitives and TVar's and tests
- The Hash array mapped trie (Hamt) datatype that is used to implement TMap and TSet.

arrow-docs/docs/fx/managing-state/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+In this example `count.update` is our atomic operation, and it is performed by possibly many threads at the same time.
+Despite that our counter increments one by one and never contains an invalid state (like say two threads incrementing 1 to 2 at the same time and both persisting 2 missing one update).
+
+### Drawbacks of using atomic variables
+
+Atomic variables have one major drawback: Their atomic guarantees do not scale beyond one variable.
+We can guarantee that access to one variable is atomic, but accessing two variables with atomic operations, even right after each other, will not make the entire operation atomic.
+
+## Securing access to shared state with locks
```
nomisRev, line 47: This can also be done with `val mutex = Semaphore(1)`, but that's a different pattern than `ConcurrentVar`.

arrow-docs/docs/fx/managing-state/README.md
```diff
@@ -0,0 +1,188 @@
...
+  (0 until 20_000).parTraverse {
+    val current = mvar.take()
+    // <1>
+    mvar.put(current + 1)
+  }
+  println(mvar.take())
+}
+```
+
+When a thread tries to invoke `ConcurrentVar.take()` it will block until the `ConcurrentVar` has a value to take and then removes the value.
```
nomisRev, line 70:
```suggestion
When a thread tries to invoke `ConcurrentVar.take()` it will suspend until the `ConcurrentVar` has a value to take and then removes the value.
```

arrow-docs/docs/fx/managing-state/README.md
```diff
@@ -0,0 +1,188 @@
...
+This means, after taking, no other thread can finish a call to `ConcurrentVar.take()` until `ConcurrentVar.put()` is called putting a value in the `ConcurrentVar`.
```
nomisRev, line 71: This is not correct. Another thread can finish a call to `ConcurrentVar.take()`, if a concurrent thread in the meanwhile calls `put`. This doesn't happen in the above example, but it's still possible.

arrow-docs/docs/fx/managing-state/README.md
```diff
@@ -0,0 +1,188 @@
...
+In our example this section is marked as `<1>`.
+
+This offers a few ways to deal with shared state now:
+- We can use one global `ConcurrentVar` to store our state and thus force all access to our state to first `take` the variable and then to `put` it back when we are done modifying.
+- We can also split our state into more fine-grained variables to allow for more concurrency.
+
+## Drawbacks
+
+A major drawback to locks is that they are tricky to use correctly:
+- Forgot to release a lock? Any future attempt at taking it will fail, and that thread will block forever!
```
nomisRev, line 82:
```suggestion
- Forgot to release a lock? Any future attempt at taking it will fail, and that thread will suspend forever!
```

arrow-docs/docs/fx/managing-state/README.md
```diff
@@ -0,0 +1,188 @@
...
+If that is not the case it will simply throw away all changes and start over.
+This effectively means that if we were to do side effects inside a transaction those may be performed many times.
+
+In most use cases this is a non issue, and the benefits of transactions far outweigh this downside.
+
+There are two other important downsides of `STM` to note:
+- Transactions are worse under high contention.
+If a variable is updated many many times a second it may be beneficial to use a different concurrency tool like atomic references or locks.
+- Transactions have no notion of fairness. If two threads wait for a lock to be released, the one who came first will get the lock next.
+This does not happen with transactions. If multiple transactions retry and block (ie they wait for a variable to change), they will all resume at the same time when a variable changes.
```
nomisRev, line 185:
```suggestion
This does not happen with transactions. If multiple transactions retry and suspends (ie they wait for a variable to change), they will all resume at the same time when a variable changes.
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TArray.kt
```diff
@@ -0,0 +1,34 @@
...
+  newTArray(size) { a }
+suspend fun <A> STM.newTArray(vararg arr: A): TArray<A> =
+  TArray(arr.map { newTVar(it) }.toTypedArray())
+suspend fun <A> STM.newTArray(xs: Iterable<A>): TArray<A> =
+  TArray(xs.map { newTVar(it) }.toTypedArray())
+
+/**
+ * A [TArray] is an array of transactional variables.
+ */
+data class TArray<A>internal constructor(internal val v: Array<TVar<A>>) {
```
nomisRev, line 17: Does this need to be a `data class`? Or can this be an `inline class`? Perhaps we should put that off until it's no longer `@Experimental`?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TMVar.kt
```diff
@@ -0,0 +1,38 @@
...
+    suspend fun <A> new(a: A): TMVar<A> = TMVar<A>(TVar.new(Option.Some(a)))
+    suspend fun <A> empty(): TMVar<A> = TMVar<A>(TVar.new(Option.None))
+  }
+}
+
+/**
+ * ADT to avoid problems with `TMVar<A?>` because kotlin can't differ `A?` from `A??` so using a nullable
+ *  reference inside `TMVar` does not work.
+ */
+sealed class Option<out A> {
```
nomisRev, line 35: There is a `Maybe` type somewhere already in Arrow Fx Coroutines. Should we promote this to `typealias Maybe<A> = Either<Unit, A>` and/or move it to the `predef.kt` at the root of the module?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TMVar.kt
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * ## Creating a [TMVar]:
+ *
+ * As usual with [STM] types there are two equal sets of operators for creating them, one that can be used inside
+ *  and one for use outside of transactions:
+ * - [TMVar.new] and [STM.newTMVar] create a new filled [TMVar]
+ * - [TMVar.empty] and [STM.newEmptyTMVar] create an empty [TMVar]
+ *
+ */
+data class TMVar<A> internal constructor(internal val v: TVar<Option<A>>) {
```
nomisRev, line 24: `inline` or wait until no longer experimental?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TQueue.kt
```diff
@@ -0,0 +1,148 @@
...
+ *  a transaction or not.
+ *
+ * ## Concurrent reading and writing
+ *
+ * [TQueue] is implemented using two [TVar]'s. One for reading and one for writing.
+ * This effectively means that writing and reading accesses two different variables and thus it never blocks each other.
+ *
+ * > In practice reads have to access the writes variable if they run out of elements to read but this is infrequent.
+ */
+data class TQueue<A> internal constructor(
```
nomisRev, line 23: Can we not wrap `IQueue` here? And or can/should we improve `IQueue` to incorporate some of these benefits? https://github.com/arrow-kt/arrow-fx/blob/master/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/IQueue.kt

1Jajen1: Not as easy no. Because for TQueue to be worth anything we need both lists to be inside `TVar`'s. And yes `IQueue` should also use the linked list implementation that we have here. It is simply better than immutable by copy arraylists. (The gc pressure produced under heavy use is insane...). Another consideration would be a finger tree for `logn` append (amortized to constant time) but that would need some benchmarks first to prove its worth.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TMVar.kt
1Jajen1: Yes in `ConcurrentVar`, I'll switch to using that :+1: We could do `Either<Unit, A>` but that means every `Left(Unit)` is a new object unless we manually create a singleton and use that `const val None = Either.Left(Unit)`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TArray.kt
1Jajen1: I had no good experience with `inlince class` yet but I'll try it :+1:

arrow-docs/docs/fx/managing-state/README.md
1Jajen1: `Semaphore(1) == ConcurrentVar(Unit)` right? Either way it has the same drawbacks

arrow-docs/docs/fx/managing-state/README.md
1Jajen1: Isn't that a terrible usage pattern? I think it is bad practice to `put` to a `ConcurrentVar` from a different section of code that has called `take` on it. I mean there is a good reason the most common use in the haskell world is `withMVar` which does all the taking and releasing in a resource safe manner.

Either way I'll try to adjust the wording a bit^^

arrow-docs/docs/fx/managing-state/README.md
nomisRev: `Semaphore(1) == ConcurrentVar(Unit)` 🤔 I have never thought about it like this 😅

I guess in this way `acquire == take` and `release` == `put`where they'er all scheduled in order. Anyway, you almost always want to use `withPermit` which is build upon bracketing `acquire` with `release` to make sure you don't lose any permits. Which is like ensuring you always correctly release your locks. (Not to be confused with JVM-level locks, which aren't used in either `ConcurrentVar` nor `Semaphore`).

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TQueue.kt
nomisRev: Oh I see! Now I'm curious to see the benchmark between using a `PList` and `copy arraylists` :D
The `IQueue` is inspired by Scala's Std `Queue`, with a small improvement over it's `size`, but there is no reason why we wouldn't want to improve it futher if we see some good opportuninties.

arrow-docs/docs/fx/managing-state/README.md
1Jajen1: Yeah `withPermit` is definitely the best pattern (haskell recommends `withMVar` over anything else as well^^). Do we also have a `withConcurrentVar` method? Could use the same treatment.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TMVar.kt
1Jajen1: Yeah control flow with nullables only works if you aren't doing it over generics, because that breaks if the generic itself is nullable...

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TArray.kt
1Jajen1: Actually thinking about this: I really do not want the constructor exposed so inline is off the table here. I really dislike the public only way that inline classes work with

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TQueue.kt
1Jajen1: I should set that up. My guess is that a normal linked list will outperform arraylists (maybe even the mutable variant) on every relevant method for queues (`prepend`, `head`, `tail`) and be on par with `reverse`. I am pretty sure that linked lists are optimal of only these 4 methods are concerned and the only thing that can beat those are mutable linked lists, but those come with its own problems. And STM and mutability simply does not work (and shouldn't).

Finger trees could work as well because they have amortized constant time `append` and `tail` (so no need for `reverse` anymore) and constant time `head` (with worst case being `O(logn)`). But the structure there is more complex and thus may be slower. Then again finger trees are amazing for concatenation so if supporting a fast `appendAll` method is desired they might be better again^^

The only thing that the STM implementation cannot do efficiently is track the size. Outside of STM this could be done by tracking appends/reads but in STM this would mean having a shared variable that both reads and writes need to go through and thus it increases contention which is poison for STM's performance. This is also the reason I don't have a size method for `TMap`/`TSet`.

arrow-docs/docs/fx/managing-state/README.md
nomisRev: No, we currently don't have such a method but that would be a great addition I think!

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TArray.kt
nomisRev: 😲 I wasn't aware of that limitation, that's a shame indeed.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TMVar.kt
nomisRev: Unless you use a nasty hack to optimize 😅 but definitely not worth it until you can benchmark and justify it the maintenance cost.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stm/TQueue.kt
nomisRev: Makes perfect sense! Thanks for clarifying further :)

1Jajen1:
I decided to remove the "managing-state" document for now. I'll probably redo that at some point, but I did not like where it was going and I don't have the time atm to start over^^

nomisRev:
> Move STM to its own module

🔥! Thank you @1Jajen1 🙏 

I am also moving Streams to a separate module, and I'll add a non-published common module for things like the internal exceptions, `UnsafePromise`, `Token` etc.

1Jajen1:
Great^^

If you are already on it: Is creating an `arrow-fx-coroutines-test` module worth it? I copied `ArrowFxSpec` into the pr after moving it, so that might be something worth sharing.

nomisRev:
> arrow-fx-coroutines-test

Yes, I think so! I was also already copying `ArrowFxSpec` & some of the helpers around.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/internal/Impl.kt
```diff
@@ -0,0 +1,214 @@
...
+  private fun merge(other: STMFrame): Unit {
+    accessMap.putAll(other.accessMap)
+  }
+}
+
+/**
+ * In some special cases it is possible to detect if a STM transaction blocks indefinitely so we can
+ *  abort here.
+ */
+object BlockedIndefinitely : Throwable("Transaction blocked indefinitely")
```
raulraja, line 163: Should these override fillInStackTrace from the Throwable constructor or is the stack trace meaningful here?

raulraja (approved):
Impressive work @1Jajen1!

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/internal/Impl.kt
1Jajen1: For this one it does not matter I think. This is a terminal exception that'll usually kill the program as there is no sensible way to recover.

I might need to do that for the internal retry exception though...

1Jajen1: Which is to say, `BlockedIndefinitely` is not internal and control flow only and thus needs the stacktrace.

I changed the other internal only exception to override `fillInStackTrace` to a noop since it is only used for control flow anyway.

Also I am using exceptions instead of suspension because I only need aborts and never resume, so it is likely more efficient which is really important here...

nomisRev (approved):
Very cool implementation of STM! 👏 👏 👏

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/internal/Hamt.kt
```diff
@@ -0,0 +1,146 @@
...
+}
+
+const val ARR_SIZE = 32 // 2^DEPTH_STEP
+const val DEPTH_STEP = 5
+const val MASK = 1.shl(DEPTH_STEP) - 1
+
+fun Int.index(): Int = MASK.and(this)
+fun Int.atDepth(d: Int): Int = shr(d)
+fun Int.indexAtDepth(d: Int): Int = atDepth(d).index()
+fun Int.nextDepth(): Int = this + DEPTH_STEP
```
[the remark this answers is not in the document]
1Jajen1: I believe this was due to usage of inline somewhere above, I'll check tomorrow if I can lower that...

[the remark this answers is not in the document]
nomisRev: Thank you @1Jajen1!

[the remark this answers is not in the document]
1Jajen1: Ah yes this is because `alterHamtWithHash` is public and inline. It has to be inline because it uses array creation methods from the stdlib which use reified afaik. This is really annoying :/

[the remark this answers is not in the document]
nomisRev: Then let's merge :)

1Jajen1:
Interesting how many new flawed bits you spot when re-reading something after a bit of time has passed :thinking: 

Either way I'll merge this later today when the build finishes^^

PR #271, opened 2020-08-26
Credit source in schedule docs

1Jajen1:
As controversial as the event leading to this may have been it is still undeniably correct that we should properly provide credit here.

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
@@ -20,6 +20,8 @@ import kotlin.random.Random
  *
  * [Schedule] allows you to define and compose powerful yet simple policies, which can be used to either repeat or retry computation.
  *
+ * > [Schedule] has been derived from scalaz zio's [Schedule](https://zio.dev/docs/datatypes/datatypes_schedule) datatype and has been adapted to kotlin.
+ *
  * The two core methods of running a schedule are:
  * - __retry__: The effect is executed once, and if it fails, it will be reattempted based on the scheduling policy passed as an argument. It will stop if the effect ever succeeds, or the policy determines it should not be reattempted again.
  * - __repeat__: The effect is executed once, and if it succeeds, it will be executed again based on the scheduling policy passed as an argument. It will stop if the effect ever fails, or the policy determines it should not be executed again. It will return the last internal state of the scheduling policy, or the error that happened running the effect.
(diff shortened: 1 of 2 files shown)


aballano:
I'm not 100% sure, but I believe it should be enough by updating the [licenses](https://github.com/arrow-kt/arrow-core/tree/master/arrow-docs/docs/legal/licenses) file in arrow-core, no? cc @raulraja

1Jajen1:
We already include scalaz's license and that is where this impl initially came from. Not sure how accurate it was back then though. But I believe the docs should mention it regardless, barely anyone looks at license files and I think it is only ethical to properly credit there. (And yes despite the unethical behavior of those credited).

raulraja (approved):

aballano:
> We already include scalaz's license and that is where this impl initially came from. Not sure how accurate it was back then though. But I believe the docs should mention it regardless, barely anyone looks at license files and I think it is only ethical to properly credit there. (And yes despite the unethical behavior of those credited).

The problem I see with this is that we might need to add more licenses in waaaaaay more files if we want to stay consistent. Otherwise we're doing an exception for this case IMO, what do you think?

1Jajen1:
> > We already include scalaz's license and that is where this impl initially came from. Not sure how accurate it was back then though. But I believe the docs should mention it regardless, barely anyone looks at license files and I think it is only ethical to properly credit there. (And yes despite the unethical behavior of those credited).
> 
> The problem I see with this is that we might need to add more licenses in waaaaaay more files if we want to stay consistent. Otherwise we're doing an exception for this case IMO, what do you think?

I don't see this as a thing about licenses at all, this is about showing (clearly and openly visible to anyone reading) where a feature is derived from.

Regarding our other docs: I actually do think that at some point we should do a pass over the docs and add links to sources in more places. This can have many forms:
If our part is very close in api and implementation to its source we should show it prominently like added here in the pr.
If our part evolved from its source and only faintly resembles it we should append it as a sort of footer.

> I don't actually think there are too many places where credit needs to be added but just to make sure :see_no_evil: 

I believe that it is only fair to other library authors (or researchers if the source is a research paper) to show it like this and I would expect the same level of credit from them.

As for licenses: Including a license is bare minimum and imo not sufficient credit if whatever was added ends up in the api visible to a user.

That being said arrow also has a lot of things that are, what I would call, "common knowledge" implementations those don't need extra credit, I am only advocating for showing this when we derive a specific feature with little (api) changes from a third-party source.

rachelcarmena (approved):
Thanks @1Jajen1 :raised_hands: We already include scalaz's license and I think it's worth to mention it in the API documentation as well. Thanks!!

rachelcarmena:
The license mentioned by @aballano was removed here: https://github.com/arrow-kt/arrow-core/pull/228 to avoid misunderstandings because it's not being updated/used but https://github.com/arrow-kt/arrow/blob/master/LICENSE :raised_hands:

PR #272, opened 2020-08-26
Add some instances to Duration and add compareTo to fx-coroutines 

1Jajen1:
After #262, fx had some functionality that fx-coroutines was missing. This adds duration comparison to fx-coroutines.

It also adds some instances so we get good coverage with laws.

At some point (probably best when kotlin time is stable) we should move Duration to core or something similar and add/test instances there to avoid duplicate code. Duration may also be useful outside of fx as well.


aballano (approved):

rachelcarmena (approved):
Thanks @1Jajen1 :raised_hands:

PR #273, opened 2020-08-28
parTupleN up to 9

tapegram:
WIP


nomisRev:
In the ticket there is some more description: 

And here is an example of how you can compose aritities 2 to get arity 4, https://github.com/arrow-kt/arrow-fx/blob/master/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt#L570.

Similarly of combining arity 2 & 3: https://github.com/arrow-kt/arrow-fx/blob/master/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt#L614

PR #275, opened 2020-08-31
Upgrade to Kotlin 1.4.0

rachelcarmena:
# Changes

* Standard library is added automatically with the plugin now
* Remove old Kotest dependency
* Remove unnecessary configuration
* Fix: Protected function call from public-API inline function is prohibited
* Disable failing tests. I've created an issue with pending tests: https://github.com/arrow-kt/arrow-fx/issues/276


danimontoya (approved):
thanks @rachelcarmena 👍

nomisRev (approved):

PR #284, opened 2020-09-22
Add internal docs cancellation

nomisRev:
Fix release called with cancelled after completion and receiving cancel signal


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
```diff
@@ -9,10 +9,10 @@ import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.startCoroutine
 
-internal open class ArrowInternalException(
-  override val message: String =
-    "Arrow-kt internal error. Please let us know and create a ticket at https://github.com/arrow-kt/arrow/issues/new/choose"
-) : RuntimeException(message)
+internal const val ArrowExceptionMessage =
```
1Jajen1, line 12: I need to use this in STM...

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/SuspendConnection.kt
```diff
@@ -26,12 +26,42 @@ inline class CancelToken(val cancel: suspend () -> Unit) {
...
+ *
+ * # DefaultConnection
+ *
+ * [DefaultConnection] keeps track of all [CancelToken] registered for a given [CoroutineContext].
+ * It does so by keeping all [CancelToken] in a FIFO stack, and running them in order when cancelled.
+ *
+ * Since [SuspendConnection] is a FIFO stack, we can register a token before running a cancellable operation,
+ * and pop the token after the cancellable operation has finished so
+ * we don't run a [CancelToken] when it's not necessary anymore since that could lead to (undefined) weird behavior.
+ * You can see this usage in [parMapN], [raceN], [racePair] & [raceTriple].
```
1Jajen1, line 63: This seems inaccurate, it may lead to some open questions like:
- Why does it being a FIFO stack mean we can register before and pop after a cancellable op? This doesn't seem related.
- When is a running CancelToken no longer necessary?

rachelcarmena (approved):
Thanks @nomisRev :tada:

1Jajen1 (approved):
Looks good to me :+1:

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
nomisRev: This can safely be non-internal 👍 

```suggestion
const val ArrowExceptionMessage =
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/SuspendConnection.kt
nomisRev: > Why does it being a FIFO stack mean we can register before and pop after a cancellable op? This doesn't seem related.

Hmm, you're correct. It's more related to how it's supposed to be used I guess. Not sure how we can phrase it more clearly.

> When is a running CancelToken no longer necessary?

For example in `parMap2`, we launch 2 tasks in parallel on an existing `SuspendConnection`.
So we create two new `SuspendConnection` for those 2 parallel tasks, and push them as a single entry on the stack of the original context.

```
val connA = SuspendConnection()
val connB = SuspendConnection()
originalContext.pushPair(connA, connB)
```

There are two scenarios where we'd want to `pop` the entry, since running the `CancelToken` is no longer necessary. When either task failed, and after we canceled the other one OR when both have completed successfully. Since in both those scenarios the tasks have already been canceled, or completed.

```
fun complete(a: A, b: B) {
  conn.pop()
  cb(
    try {
      Result.success(f(a, b))
    } catch (e: Throwable) {
      Result.failure<C>(e.nonFatalOrThrow())
    }
  )
}
fun sendException(other: SuspendConnection, e: Throwable) = when (state.getAndSet(e)) {
  is Throwable -> Unit // Do nothing we already finished
  else -> other.cancelToken().cancel.startCoroutine(Continuation(EmptyCoroutineContext) { r ->
    conn.pop()
    cb(Result.failure(r.fold({ e }, { e2 -> Platform.composeErrors(e, e2) })))
  })
}
```

raulraja (approved):
Looks great, thanks Simon!

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/SuspendConnection.kt
nomisRev: @1Jajen1 any thoughts on how we an improve?

1Jajen1: Not really, I think this explanation on when `CancelToken`'s are no longer necessary is good to have, but for the stack, I have no idea how to phrase it better :see_no_evil:

nomisRev: No problem, I'll give this a shot to re-phrase better with the added example.

PR #286, opened 2020-09-25
Result doesn't need to be nullable

pablisco:
In the terminal operation `firstOrError` we return `O?`. However, there is possible way to return null as we throw an instance of `NoSuchElementException`.


nomisRev (approved):
Thanks for the fix @pablisco!!

raulraja (approved):
thanks!

nomisRev:
@pablisco I think it's best you pushed again to retrigger Actions with the changed workflow.

PR #288, opened 2020-09-26
Address `Either.catch` inline incompatibility.

joramnv:
This PR relates to the change to inline the `Either.catch` which is in this PR: https://github.com/arrow-kt/arrow-core/pull/237.

Somehow, the compiler is not happy with the current usage at line 238 in combination with the inlined version of `Either.catch`. This PR addresses that issue by incorporating the `Either.catch` functionality in the internal `go` function.

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/Compiler.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/Compiler.kt
@@ -4,6 +4,9 @@ import arrow.core.Either
 import arrow.core.None
 import arrow.core.Some
 import arrow.core.getOrElse
+import arrow.core.left
+import arrow.core.nonFatalOrThrow
+import arrow.core.right
 import arrow.fx.coroutines.ExitCase
 import arrow.fx.coroutines.Platform
 import arrow.fx.coroutines.Token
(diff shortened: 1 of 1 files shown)


rachelcarmena:
Great @joramnv, thank you so much!! :raised_hands: 

Every log shows now:

```
Looking for joram-either-resolve in arrow-core ...
joramnv:joram-either-resolve found for arrow-core!
From https://github.com/joramnv/arrow-core
 * branch            joram-either-resolve -> FETCH_HEAD
Updating 244cb7d..995f01c
Fast-forward
.../src/main/kotlin/arrow/core/Either.kt           | 143 +++++++++++++++++++-
.../src/test/kotlin/arrow/core/EitherTest.kt       | 144 +++++++++++++++++++++
.../arrow/core/test/generators/Generators.kt       |  40 ++++++
3 files changed, 326 insertions(+), 1 deletion(-)
```

so both PRs are _playing_ at the same time.

It seems Arrow Fx continues raising the same error. I'll add the support for HTTPS very soon so you will be able to check it locally as well. Meanwhile, you can use these checks. Thanks again!!

raulraja (approved):
thanks @joramnv !

nomisRev (approved):
Thanks for this @joramnv!!

PR #289, opened 2020-09-29
Docs abouts internals Arrow Fx Coroutines + integration IO

### LINKED ISSUE #176: ["Request"] Integrate Arrow Fx Coroutines into `IO`
Integrate the cancellation system of Arrow Fx Coroutines to `IO`.

This will allow automatically integrate with the new `suspend` code and `IO` code when using the `0.11.0` version of `IO`, which will come without breaking API changes.

nomisRev:
Adds a document about the internals of Arrow Fx Coroutines, and how it's cancellation system works on top of Kotlin's Coroutine system. More docs about `uncancellable`, `bracket` or parallel ops can follow.

Moves SuspendConnection to a seperate module, and shares its internals with IO's cancellation system.
This automatically provides cancellation integration between the two, and shows how the cancellation system of Arrow Fx Coroutines is the same as that of IO. So this closes #176.


nomisRev:
@rachelcarmena I have some issues with the build that I'm not sure about.

The tests of Arrow Fx Coroutines fail since it says it cannot find the code from `arrow-fx-suspend-connection` but it's included as an `implementation` dependency 🤔 

Also `arrow-fx-coroutines-kotlinx-coroutines` fails with errors in animal sniffer due to not being able to find `SuspendConnection` related code, but there is no change in that module at all.

Any ideas? Did I apply `implementation` incorrectly? The idea was that both Arrow Fx & Arrow Fx Coroutines internally depend on `SuspendConnection` but it's not exposed to the user.

rachelcarmena:
Hi @nomisRev , `implementation`  is exposed as runtime. Please, take a look at `test...`  configurations: https://github.com/arrow-kt/arrow/blob/master/docs/gradle/configuration.md

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
 # Arrow Fx Coroutines
 
-Arrow Fx Coroutines offers a suspend DSL with the same semantics and API you find in Arrow Fx's IO, this allows for writing programs using suspend fun directly instead of wrapping in IO.
-It's a functional implementation of Kotlin's coroutine system while providing support for cancellation, error handling, resource handling and all goodies you might be familiar with from other functional effect systems.
+[![Arrow Fx logo](https://raw.githubusercontent.com/arrow-kt/arrow-site/master/docs/img/fx/arrow-fx-brand-sidebar.svg?sanitize=true)](https://arrow-kt.io)
 
-Arrow Fx aims to be a battery included functional effects framework, below you can find a small overview of the library.
-A full coverage of all data types and functions can be found [here](LINK DOCS).
+Λrrow Fx is part of Arrow Fx [**Λrrow**](https://arrow-kt.io/docs/fx/), and the public documentation and information on how to use the library can e found there.
```
raulraja, line 5:
```suggestion
Λrrow Fx is part of Arrow Fx [**Λrrow**](https://arrow-kt.io/docs/fx/), and the public documentation and information on how to use the library can be found there.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
+```kotlin
+interface Continuation<A> {
+	val ctx: CoroutineContext
+	fun resume(result: Result<A>)
+}
+```
 
-suspend fun fetchUser(): Either<PersistenceError, User> =
-    Right(User("simon@arrow-kt.io", "Simon"))
+As mentioned above, a `suspend` programs can return immediately or `suspend` and return it's value at aa later point in time. Thus this can be modeled with a `Contination<A>` where both the immediate result or a suspended result is always returned to `resume`, this ensures that the result is always consumed in the same manner no matter how the `suspend` program behaves.
```
raulraja, line 43:
```suggestion
As mentioned above, a `suspend` programs can return immediately or `suspend` and return it's value at a later point in time. Thus this can be modeled with a `Contination<A>` where both the immediate result or a suspended result is always returned to `resume`, this ensures that the result is always consumed in the same manner no matter how the `suspend` program behaves.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
+### startCoroutine
 
-// Or unwrapped in `suspend`
-suspend suspendedIOProgram(): Either<PersistenceError, ProcessedUser> =
-  ioProgram().suspended()
- ```
+The most high level runner in the standard library is `startCoroutine`, and it's the first method we'll discuss that creates a `Coroutine` and runs it. Kotlin’s standard library defines a Coroutine as an instance of a suspendable computation. In other words, a Coroutine is a compiled `suspend () -> T` program wired to a Continuation.
 
-### suspend () -> Either<E, A>
+So let’s take a quick look at an example with the 3 types we've seen and discussed above.
```
raulraja, line 55:
```suggestion
Let’s take a quick look at an example with the 3 types we've seen and discussed above.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
 ```
 
-## Cancellation
+Here we can see that `ContinuationInterceptor` is a `CoroutineContext.Element`, which means we can put into an immutable `CoroutineContext` map. It has an abstract `interceptContinuation` method, which is called upon calling `intercepted`. It can be called upon the `Continuation` returned by `createCoroutineUnintercepted`, and in that case will call `interceptContinuation` in the case the `CoroutineContext` contains a `ContinuationInterceptor`. This operation is idempotent, since the result is cached.
 
-The cancellation system exists out of a few simple building blocks.
+When the `Continuation` is finished running, then before returning it will call `releaseInterceptedContinuation`, but only in the case `interceptContinuation` returned a different `Continuation` than it received.
 
-All operators found in Arrow Fx check for cancellation. In the small example of an infinite sleeping loop below `sleep` checks for cancellation and thus this function also check for cancellation before/and while sleeping.
+So let's take a look at a simple example.
```
raulraja, line 111:
```suggestion
Let's take a look at a simple example.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
+Secondly, our `Printer` continuation is started with `resume(Unit)`, which we see reflected as receiving `Success(kotlin.Unit)` in `resume` and then simply passing it along to our original continuation. The original continuation, created by the Kotlin Compiler, will then start our `helloWorld` suspend program.
 
-2. When the resulting `suspend` operation is cancelled than all running fibers inside will also be cancelled so that all paralell running task can gracefully exit and close their resources before returning.
+3. After our program is finished, we see that our `Printer` continuation is released.
 
-For more documentation on parallel operations see below.
+4. Finally, we see that our result is finally returned to our own `Continuation`.
 
-### Uncancellable
+So you might be wondering why this is useful? This isn't very useful for printing anything obviously, because we only get `intercepted` before anything starts running and released when it's finished. But it gives us the perfect entry point to schedule the coroutine on a `Thread` or `ExecutorService`.
```
raulraja, line 170: nice transition 👏

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
+  println("Fired Twice")
+
+  cont.resume(Unit)
+  println("Fired three times")
+}
 ```
 
-The difference between `Resource` is that `bracketCase` is simple function, while `Resource` is a data type, both ensure that resources are `acquire`d and `release`d correctly.
-It also forms a `Monad` so you can use it to safely compose `Resource`s, map them or safely traverse `Resource`s.
+So if we look back at `startCoroutine`, we see that is simply creates a Coroutine, **always** invokes its interceptor and then immediately starts it. So the highest level runner, combines the lowest level functions.
```
raulraja, line 264:
```suggestion
If we look back at `startCoroutine`, we see that is simply creates a Coroutine, **always** invokes its interceptor and then immediately starts it. So the highest level runner, combines the lowest level functions. 
```

raulraja (approved):
Thanks so much @nomisRev ! 🙏 .

nomisRev:
Thanks for the review and feedback @raulraja 🙏

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
-To ensure resource safety we need to take care with cancellation since we don't wont our process to be cancelled but our resources to remain open.
+### createCoroutineUnintercepted
 
-There Arrow Fx offers 2 tools `Resource` and `suspend fun bracketCase`. Any `resource` operations exists out of 3 steps.
+`createCoroutineUnintercepted` is the lowest level function of creating (and thus subsequently starting) a Coroutine, and you'll find it is used in most other function with either `createCoroutineXXX` or `startCoroutineXXX`.
 
-1. Acquiring the resource
-2. Using the resource
-3. Releasing the resource with either a result, a `Throwable` or `Cancellation`.
+It's found in the `kotlin.coroutines.intrinsics` package, which we saw 1 method of so far `intercepted`. And we'll cover the other content below. As we've already mentioned above, this method turns a `suspend () -> A` function into a `Coroutine`, which is an instance of a suspendable computation. This Coroutine hasn't started running yet, and we can run it as many times as we want. As you might've guessed from above, we can start the Coroutine by calling `resume(Unit)` on it.
```
danimontoya, line 235: For consistency with aboves
```suggestion
It's found in the `kotlin.coroutines.intrinsics` package, which we saw 1 method of so far `intercepted`. And we'll cover the other content below. As we've already mentioned above, this method turns a `suspend () -> T` function into a `Coroutine`, which is an instance of a suspendable computation. This Coroutine hasn't started running yet, and we can run it as many times as we want. As you might've guessed from above, we can start the Coroutine by calling `resume(Unit)` on it.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
 
-To ensure the resource can be correctly acquired we make the `acquire` & `release` step `uncancelable`.
-If the `bracketCase` was cancelled during `acquire` it'll immediately go to `release`, skipping the `use` step.
+For a given program `suspend () -> A` we pass a `Continuation<A>`, and `createCoroutineUnintercepted` returns us a `Continuation<Unit>`. This `Continuation<Unit>` models a Coroutine, which is the program `suspend () -> A` wired to it's `context` and callback `resume: (Result<A>) -> Unit`.
```
danimontoya, line 237:
```suggestion
For a given program `suspend () -> T` we pass a `Continuation<A>`, and `createCoroutineUnintercepted` returns us a `Continuation<Unit>`. This `Continuation<Unit>` models a Coroutine, which is the program `suspend () -> T` wired to it's `context` and callback `resume: (Result<T>) -> Unit`.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
-  }
+```
+public fun <T> (suspend () -> T).startCoroutine(completion: Continuation<T>) {
+    createCoroutineUnintercepted(completion).intercepted().resume(Unit)
 }
+```
+
+### suspendCoroutineUninterceptedOrReturn / COROUTINE_SUSPENDED
+
+As you might've noticed in all examples above, we haven't actually used any suspension so far. But what does it mean to suspend? Most simply put it means the function doesn't immediately returns, but is going to return the result through the a callback `(Result<A>) -> Unit`.
```
danimontoya, line 274:
```suggestion
As you might've noticed in all examples above, we haven't actually used any suspension so far. But what does it mean to suspend? Most simply put it means the function doesn't immediately returns, but is going to return the result through the callback `(Result<T>) -> Unit`.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
+```
+Immediately returned: Hello World!
+I am called with result: Success(Hello World!)
+suspending returned: COROUTINE_SUSPENDED
+```
 
-## Error Handling
+Here in the output we can clearly see that in the first case the callback is never invoked, and we get the result immediately returned to us.
+In contrast to the second case where the callback is invoked, **and** the immediate returned value is `COROUTINE_SUSPENDED`.
+This means that if we use `startCoroutineUninterceptedOrReturn` you **always** need to check the returned value to check if it's `COROUTINE_SUSPENDED` or a user returned value (which in that case needs to be casted back from Any? to A).
```
danimontoya, line 384:
```suggestion
This means that if we use `startCoroutineUninterceptedOrReturn` you **always** need to check the returned value to check if it's `COROUTINE_SUSPENDED` or a user returned value (which in that case needs to be casted back from Any? to T).
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
+  suspendCoroutineUninterceptedOrReturn { cont ->
+    if (cont.context[SuspendConnection]?.isCancelled() == true) COROUTINE_SUSPENDED
+    else Unit
   }
-}
 ```
 
-## Parallel Ops
+So a `cancelBoundary` checks if the `SuspendConnection#isCancelled` and if so it suspends forever and never invokes the callback, effectively this means it stops running the code.
+In the case the `SuspendConnection` isnt' cancelled it immediately returns `Unit`. Here we use `suspendCoroutineUninterceptedOrReturn` since we can return immediately when no cancellation signal was triggered, and we don't want to intercept or schedule but we have to check the `Continuation#context` to be able to access the `SuspendConnection`.
```
danimontoya, line 441:
```suggestion
In the case the `SuspendConnection` isn't cancelled, it immediately returns `Unit`. Here we use `suspendCoroutineUninterceptedOrReturn` since we can return immediately when no cancellation signal was triggered, and we don't want to intercept or schedule but we have to check the `Continuation#context` to be able to access the `SuspendConnection`.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
 
+So this is how from any user or library code we can add checks for cancellation that automatically exit when a cancellation signal was trigerred.
 
-### ForkConnected
+When `isCancelled` returns `true`, this means that all the registered `CancelToken`'s inside `SuspendConnection` have run, or are running.
 
-`forkConnected` launches a `suspend () -> A` function on a given `CoroutineContext` inside a `Fiber`, the lifecycle of this `Fiber` is connected to its parent.
-If the parent cancels that this operation will also be cancelled.
+NOTE:
+In Kotlin 1.4, CancellationException was added to the Kotlin Std as an ExperimentalApi. The documentation states that a cancellable coroutine should throw it to signal that it has been canccelled. In a future version of Arrow Fx Coroutines we can follow this pattern by instead of suspend upon cancellation like we did in Arrow Fx IO, and as is known from Cats Effects IO, we can `throw CancellationException`. This would also only happen when we're running on a cancellable `SuspendConnection`, and we'd still be able to control cancellation in the same way as we know now are you might be familiar with from Cats Effects IO or Arrow Fx IO.
```
danimontoya, line 448:
```suggestion
In Kotlin 1.4, CancellationException was added to the Kotlin Std as an ExperimentalApi. The documentation states that a cancellable coroutine should throw it to signal that it has been canccelled. In a future version of Arrow Fx Coroutines we can follow this pattern by instead of suspend upon cancellation like we did in Arrow Fx IO, and as is known from Cats Effects IO, we can `throw CancellationException`. This would also only happen when we're running on a cancellable `SuspendConnection`, and we'd still be able to control cancellation in the same way as you might be familiar with from Cats Effects IO or Arrow Fx IO.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
 
-A `Fiber` is a data type that represents a `suspend fun join(): A` and `suspend fun cancel(): Unit` handle.
+### startCoroutineCancellable
 
-### ForkScoped
+So how do we start a Coroutine with `SuspendConnection`, so we can start `suspend () -> A` in a cancellable way. Using `SuspendConnection` is no different than using any other `CoroutineContext`, so let's take a look at the most simple example we can come up with.
```
danimontoya, line 452:
```suggestion
So how do we start a Coroutine with `SuspendConnection`, so we can start `suspend () -> T` in a cancellable way. Using `SuspendConnection` is no different than using any other `CoroutineContext`, so let's take a look at the most simple example we can come up with.
```

arrow-fx-coroutines/README.MD
```diff
@@ -1,561 +1,561 @@
...
+      CancelToken.unit
     }
-  )
+    
+    cancellable.complete(token)
+  }
 ```
 
-## And more…
+We `suspendCoroutine` to access the `Continuation`, and since we use `suspendCoroutine` here it will also `interept` and thus behave in the same safe way as Kotlin's high-level combinators.
```
danimontoya, line 534:
```suggestion
We `suspendCoroutine` to access the `Continuation`, and since we use `suspendCoroutine` here it will also `intercept` and thus behave in the same safe way as Kotlin's high-level combinators.
```

danimontoya (approved):
Very well written @nomisRev 🚀 ! Thank youu!! Only added a few nits while reading through!

nomisRev:
Thank you for the review @danimontoya!

nomisRev:
@rachelcarmena seemed unrelated to `implementation`, but rather `implementation` code being `inline` so the consumer library had to `inline` code that was only available as `implementation`.

PR #291, opened 2020-09-29
Add alternative strategy for first/last terminal ops and created tests for Terminal operations

pablisco:
The aim of this PR is to add the option to have an alternative value or exception when first or last terminal operations are invoked and it's not possible to obtain them:

The following functions:
```kotlin
Stream.empty<Int>().firstOrError() // result in NoSuchElementException
Stream.empty<Int>().lastOrError() // result in NoSuchElementException
```

Are now refactored to:
```kotlin
Stream.empty<Int>().firstOrElse { error("Oops!") } // result in IllegalStateException("Oops!")
Stream.empty<Int>().lastOrElse { error("Oops!") // result in IllegalStateException("Oops!")
// or
Stream.empty<Int>().firstOrElse { -1 } // result in -1
Stream.empty<Int>().lastOrElse { -1 } // result in -1
```

Also, we are renaming `firstOrNull()` to `first()` and `lastOrNull()` to `last()`. It doesn't class with the non blocking operators `first { }` and `last { }` as they require a condition to filter (arguably filter may be enough).

It also adds tests for all terminal operations available.


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -106,3 +134,6 @@ suspend fun <O, B> Stream<O>.foldChunks(init: B, f: (B, Chunk<O>) -> B): B =
 
 private suspend fun <O, B> Stream<O>.compiler(init: B, foldChunk: (B, Chunk<O>) -> B): B =
   asPull.compiler(init, foldChunk)
+
+@Suppress("NOTHING_TO_INLINE") // we want to avoid polluting the stack trace
```
nomisRev, line 138: Does this work to not pollute the stack trace?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -84,14 +97,29 @@ suspend fun <O> Stream<O>.lastOrNull(): O? =
...
  */
-suspend fun <O> Stream<O>.lastOrError(): O =
-  lastOrNull() ?: throw NoSuchElementException()
+suspend fun <O> Stream<O>.lastOrError(failStrategy: FailStrategy = ::noSuchElement): O =
+  lastOrNull() ?: failStrategy()
+
+/**
+ * This typealias defines a function that can only compile if we throw an exception, or call a function that returns [Nothing].
+ */
+typealias FailStrategy = () -> Nothing
```
nomisRev, line 122: Nit: Seems a bit excessive to create a public `typealias` for this, since we'll expose that `typealias` to all user codebases.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/stream/TerminalOpsTest.kt
```diff
@@ -0,0 +1,148 @@
...
+
+import arrow.fx.coroutines.Duration
+import arrow.fx.coroutines.StreamSpec
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.throwable.shouldHaveMessage
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+class TerminalOpsTest : StreamSpec(spec = {
```
nomisRev, line 11: Thank you @pablisco! ❤️

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -84,14 +97,29 @@ suspend fun <O> Stream<O>.lastOrNull(): O? =
...
 /**
- * Runs all the effects of this [Stream], raising a [NoSuchElementException] if the stream emitted no values,
+ * Runs all the effects of this [Stream], raising a [NoSuchElementException], by default, if the stream emitted no values,
  * and returning the last value emitted otherwise.
  *
  * This a terminal operator, meaning this functions `suspend`s until the [Stream] finishes.
  * If any errors are raised while streaming, it's thrown from this `suspend` scope.
+ *
+ * It's possible to include an alternative [FailStrategy] which describes how to fail in case of not finding the desired element.
+ * //sampleStart
```
nomisRev, line 107:
```suggestion
 * It's possible to include an alternative [FailStrategy] which describes how to fail in case of not finding the desired element.
 *
 * ```kotlin:ank:playground
 * //sampleStart
```

nomisRev:
> It also adds tests for all terminal operations available.

❤️

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
1Jajen1: Also I think this may be better modeled as `() -> O` and treated as a general fallback rather than an error. This allows a bit more use for those that would directly go to a fallback or catching anyway. `throw` would still work just as well because `Nothing` can be used in place of any type.

pablisco: I was thinking about that while doing this... We could do something like:
```
fun Stream<O>.firstOrElse(block: suspend () -> O)
```
And I would go even further that if we rename `firstOrNull()` to `first()` and we create the nullable operators that match that of Option then we can do:
```kotlin
stream.first().orElse { error("Boom!") }
```

I've been debating with myself whether we want to encourage composition of operations (as good functional programmers) or provide shortcuts like these that do more than one thing.  I guess we could have both and let users decide what to use :) I'm leaning more on the more atomic operations :)

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
pablisco: I think that, from memory, inline functions don't show up on the stack trace... Maybe should add some tests for this 🤔

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -47,7 +47,7 @@ suspend fun <O> Stream<O>.drain(): Unit =
  * This a terminal operator, meaning this functions `suspend`s until the [Stream] finishes.
  * If any errors are raised while streaming, it's thrown from this `suspend` scope.
  */
-suspend fun <O> Stream<O>.firstOrNull(): O? =
+suspend fun <O> Stream<O>.first(): O? =
```
nomisRev, line 50: This is inconsistent with Kotlin's standard library naming.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -93,7 +93,7 @@ suspend fun <O> Stream<O>.firstOrError(failStrategy: FailStrategy = ::noSuchElem
  * This a terminal operator, meaning this functions `suspend`s until the [Stream] finishes.
  * If any errors are raised while streaming, it's thrown from this `suspend` scope.
  */
-suspend fun <O> Stream<O>.lastOrNull(): O? =
+suspend fun <O> Stream<O>.last(): O? =
```
nomisRev, line 96: This is inconsistent with Kotlin's standard library naming.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
pablisco: You are right :) I was thinking if we are to add an extension to nullables like this:
```kotlin
fun <A> A?.orElse(f: () -> A): A = this ?: f()
```
Then, we wouldn't need to have `firstOrElse { }` and we could do: `first().orElse {  }`

I think Kotlin added `firstOrNull()` I think is because they already had `first()` that throws an error if there is no first element of the relevant collection... I think, from the get-go they should have returned a nullable value (and I think they probably would too) as it's safer.


All that said, I've made it consistent with stdlib :)

nomisRev: > I think, from the get-go they should have returned a nullable value (and I think they probably would too) as it's safer.

I agree that there should be no `first` that throws an exception! but I think Kotlin added both consciously. Not 100% sure, but `require` and `check` is also often promoted by Kotlin.

> All that said, I've made it consistent with stdlib :)
👍 🎉

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -58,9 +58,22 @@ suspend fun <O> Stream<O>.firstOrNull(): O? =
...
+ * suspend fun main(): Unit =
+ *   Stream.empty<Int>()
+ *     .lastOrElse { error("Oops!") } // equivalent to .lastOrError { throw IllegalStateException("Oops!") }
+ *     .let(::println) // not reached, IllegalStateException("Oops!") is thrown
+ * //sampleEnd
+ * ```
  */
-suspend fun <O> Stream<O>.firstOrError(): O =
-  firstOrNull() ?: throw NoSuchElementException()
+suspend fun <O> Stream<O>.firstOrElse(alternative: () -> O): O =
```
nomisRev, line 75: I think `alternative` can now also be `suspend`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -84,14 +97,28 @@ suspend fun <O> Stream<O>.lastOrNull(): O? =
...
+ *   Stream(1, 2, 3)
+ *     .drop(3)
+ *     .lastOrElse { error("Oops!") } // equivalent to .lastOrError { throw IllegalStateException("Oops!") }
+ *     .let(::println) // not reached, IllegalStateException("Oops!") is thrown
+ * //sampleEnd
+ * ```
  */
-suspend fun <O> Stream<O>.lastOrError(): O =
-  lastOrNull() ?: throw NoSuchElementException()
+suspend fun <O> Stream<O>.lastOrElse(alternative: () -> O): O =
```
nomisRev, line 120: alternative could be `suspend`.

nomisRev (commented):
This PR is shaping up nicely! The docs need another update since they still reflect the old documentation.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -58,9 +58,22 @@ suspend fun <O> Stream<O>.firstOrNull(): O? =
...
+ * suspend fun main(): Unit =
+ *   Stream.empty<Int>()
+ *     .firstOrElse { error("Oops!") } // equivalent to .firstOrElse { throw IllegalStateException("Oops!") }
+ *     .let(::println) // not reached, IllegalStateException("Oops!") is thrown
+ * //sampleEnd
+ * ```
  */
-suspend fun <O> Stream<O>.firstOrError(): O =
-  firstOrNull() ?: throw NoSuchElementException()
+suspend fun <O> Stream<O>.firstOrElse(alternative: suspend () -> O): O =
```
1Jajen1, line 75: Is there a difference between having `alternative` be `suspend` and the whole function being `inline` instead?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/TerminalOps.kt
```diff
@@ -84,14 +97,28 @@ suspend fun <O> Stream<O>.lastOrNull(): O? =
   foldChunks<O, O?>(null) { acc, c -> c.lastOrNull() ?: acc }
 
 /**
- * Runs all the effects of this [Stream], raising a [NoSuchElementException] if the stream emitted no values,
+ * Runs all the effects of this [Stream], raising a [NoSuchElementException], by default, if the stream emitted no values,
```
1Jajen1, line 100: There is no default impl atm right?

1Jajen1 (approved):
Looks good :+1:

nomisRev:
This became obsolute with the deprecation of `Stream` in favor of `Flow`.
We might add additional terminal operators as extension functions on `Flow`.

pablisco:
Following up here: https://github.com/arrow-kt/arrow-fx/issues/325

PR #292, opened 2020-09-29
[WIP] Stream test tool

pablisco:
This PR adds a new module to help test streams.

The main entry point of this API is the function `testStream { ... }`. The provided closure has a `TestStreamScope` receiver. 

`TestStreamScope` defines the following operations available to work with streams in tests:

 - capture
 - captureInOrder
 - expect
 - expectThat
 - expectAll
 - expectInOrder
 - expectException
 - expectNothingMore
 - next
 - nextOrNull
 - nextException
 - next(n)


arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
```diff
@@ -0,0 +1,81 @@
...
+  suspend fun nextException(): Any? = exceptionQueue.next()
+
+  suspend fun next(n: Int) =
+    queue.dequeue().timeout(timeout).take(n).toList().toTypedArray()
+
+  private val queue = Queue.unsafeUnbounded<Any?>()
+  private val exceptionQueue = Queue.unsafeUnbounded<Throwable>()
+
+  private suspend fun <O> Queue<O>.next(): O =
+    dequeue().interruptAfter(timeout).firstOrError { "Timeout after $timeout" }
```
nomisRev, line 69:
```suggestion
    timeOutOrNull(timeout) { dequeue1() } ?: error("Timeout after $timeout")
```

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
```diff
@@ -0,0 +1,81 @@
...
+@ExperimentalTime
+fun testStream(
+  timeout: Duration = 1.toDuration(DurationUnit.SECONDS),
+  block: suspend TestStream.() -> Unit
+): Unit = testStreamCompat(timeout = timeout.toArrowDuration(), block = block)
+
+fun testStreamCompat(
+  timeout: ArrowDuration = ArrowDuration(1, TimeUnit.SECONDS),
+  block: suspend TestStream.() -> Unit
+): Unit = runBlocking { TestStream(timeout = timeout).block() }
```
nomisRev, line 29:
```suggestion
): Unit = arrow.fx.coroutines.Enviroment().unsafeRunSync { TestStream(timeout = timeout).block() }
```

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
```diff
@@ -0,0 +1,81 @@
...
+
+// TODO: Convert to Kotlin Duration when not experimental
+class TestStream internal constructor(private val timeout: ArrowDuration) {
+
+  fun Stream<*>.capture() {
+    GlobalScope.launch {
+      runCatching { through(queue.enqueue()).drain() }
+        .onFailure { exceptionQueue.enqueue1(it) }
+    }
```
nomisRev, line 38:
```suggestion
    arrow.fx.coroutines.Enviroment().unsafeRunAsnyc {
      through(queue.enqueue())
        .handleErrorWith { Stream.effect { exceptionQueue.enqueue1(it) } }
        .drain()
    }
```

nomisRev (commented):
Not a 100% sure why "// TODO: Not working ATM" is not working, but afaik `GlobalScope` is unsafe for this, no?

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
```diff
@@ -0,0 +1,81 @@
...
+    check(next.contentEquals(items)) { "Expected $items but got $next" }
+  }
+
+  suspend inline fun expectException(exception: Throwable): Unit = nextException().let { next ->
+    check(next == exception) { "Expected $exception but got $next" }
+  }
+
+  suspend fun expectNothingMore() {
+    val values = queue.dequeue().interruptAfter(now).toList()
+    check(values.isEmpty()) { "Expected nothing more but got: $values" }
```
nomisRev, line 55:
```suggestion
    val option = queue.tryDequeue1()
    check(option.isEmpty()) { "Expected nothing more but got: $values" }
```

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
pablisco: This change makes `expects multiple items` fail :(

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
pablisco: Hmm, this one makes the "expect items in order" test hang and not timeout or success

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
pablisco: Interesting, this change makes the test that is failing work... however, I feel that we may be losing potential data... My intention was to basically flush the queue of all current values. So we can tell the tester that there were one or more items that were emitted...

pablisco: Never mind, the success was a fluke 🙃  Damn concurrency is hard... need to find a better way to control the coroutines in tests

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStream.kt
pablisco: Damn it, I typed Sync instead of Async... in it goes :)

pablisco:
@nomisRev It should be fully working now with more options available :) I need to add some extra documentation.

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStreamScope.kt
```diff
@@ -0,0 +1,177 @@
...
+   *
+   * @see capture
+   */
+  fun captureInOrder(vararg streams: Stream<*>) {
+    Environment().unsafeRunAsync {
+      Stream(*streams).effectMap { it.blockingCapture() }.flatten().drain()
+    }
+  }
+
+  private suspend fun Stream<*>.blockingCapture(): Stream<Unit> =
```
nomisRev, line 70:
```suggestion
  private fun Stream<*>.blockingCapture(): Stream<Unit> =
```

arrow-fx-coroutines-test/src/main/kotlin/arrow/fx/coroutines/stream/test/TestStreamScope.kt
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  /**
+   * Captures all the values emitted by each of the provided streams in order.
+   *
+   * @see capture
+   */
+  fun captureInOrder(vararg streams: Stream<*>) {
+    Environment().unsafeRunAsync {
+      Stream(*streams).effectMap { it.blockingCapture() }.flatten().drain()
```
nomisRev, line 66:
```suggestion
      Stream(*streams).flatMap { it.blockingCapture() }.drain()
```

nomisRev (commented):
What benefits would this library offer over using the build in `Stream` operators?

Most of the tests in `TestStreamTest` can be written in a simpler manner using simple terminal operators like done in the test suite of `Stream` itself.

pablisco:
@nomisRev 
> What benefits would this library offer over using the build in Stream operators?

For simple cases, I think it's clear that it's simpler to use terminal operations. I think it adds value when we start looking at more complex workflows as it's easier to lay down expectations in a linear way... I guess could call it test comprehensions :) 

I started this idea while testing view bindings with instrumentation tests. Something like:

```kotlin
testStream {
  onActivity { activity -> activity.button.clicks().capture() } // runs on the main android Looper
  onView(withId(R.id.button)).perform(click())
  expect(Unit)
  onView(withId(R.id.button)).perform(click())
  expect(Unit)
  expectNothingMore()
}
```

In this case, I couldn't extract the stream of clicks from the button from the main thread and I wasn't able to run espresso operations on the main thread.

nomisRev:
@pablisco thanks for looking into this. This became obsolete by deprecating Stream in favor of Flow from KotlinX Coroutines. We'd like to offer a set of extension functions over `Flow` that offer the same APIs as we have with `Stream`.

pablisco:
Totally understand. Makes sense. I may try to do something for flow if I find the time as I've been thinking about such extensions for a while. Closing this for now though 👍

nomisRev:
@pablisco I'll try to add some tickets on the repo with ideas for `Flow` extensions feel free to do the same.

PR #296, opened 2020-10-02
Make Stream.callback run on the current thread

### LINKED ISSUE #295: Stream.cancellable always runs in ComputationPool

0.11.0

What would you like to see?

It'd be desirable to be able to pass specific `CoroutineContext` to `Stream.callback` and `Stream.cancellable`, given those always run on the `ComputationPool` now (via `ForkAndForget` in the `bracketCase`) and that doesn't work in Android for wrapping Android callbacks. Android callbacks are called in the UI thread the big majority of the time, and many times you get compile time errors by the system like this when trying to create the `Stream.cancellable`:

```
...
 Caused by: java.lang.IllegalStateException: Method addObserver must be called on the main thread
```

A usual example to reproduce this:

```kotlin
fun Fragment.lifecycleDestroy(): Stream<Boolean> = Stream.cancellable {
  val lifecycleObserver = LifecycleEventObserver { _, event ->
    if (event == Lifecycle.Event.ON_DESTROY) {
      emit(true)
    }
  }

  viewLifecycleOwner.lifecycle.addObserver(lifecycleObserver)
  CancelToken { viewLifecycleOwner.lifecycle.removeObserver(lifecycleObserver) }
}
```

aballano:
Try to fix #295

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
@@ -3,7 +3,6 @@ package arrow.fx.coroutines.stream
 import arrow.core.Either
 import arrow.fx.coroutines.CancelToken
 import arrow.fx.coroutines.ExitCase
-import arrow.fx.coroutines.ForkAndForget
 import arrow.fx.coroutines.Promise
 import arrow.fx.coroutines.UnsafePromise
 import arrow.fx.coroutines.andThen
@@ -121,7 +120,7 @@ fun <A> Stream.Companion.cancellable(@BuilderInference f: suspend EmitterSyntax<
     val cancel = Promise<CancelToken>()
 
     Stream.bracketCase({
-      ForkAndForget(EmptyCoroutineContext) { emitterCallback(f, cancel, error, q) }
+      emitterCallback(f, cancel, error, q)
     }, { f, exit ->
       when (exit) {
         is ExitCase.Cancelled -> cancel.get().cancel.invoke()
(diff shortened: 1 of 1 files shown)


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -121,7 +121,7 @@ fun <A> Stream.Companion.cancellable(@BuilderInference f: suspend EmitterSyntax<
     val cancel = Promise<CancelToken>()
 
     Stream.bracketCase({
-      ForkAndForget { emitterCallback(f, cancel, error, q) }
+      ForkAndForget(EmptyCoroutineContext) { emitterCallback(f, cancel, error, q) }
```
JorgeCastilloPrz, line 124: What's the reason to fork here if we staying on the same context? 🤔

nomisRev: `Fork`'ing on `EmptyCoroutineContext` doesn't actually fork, and the concern is still left to the user like @JorgeCastilloPrz linked in the snippet of the issue? In that case, there is no need to `ForkAndForget` here.

`lifecycle.addObserver` is the same as the use-case of a websocket.

JorgeCastilloPrz: The only difference is that Android will crash if it detects you attach the observer in a thread that's not the Main one

aballano: FYI tests failed after removing `ForkAndForget`, I'm unsure if it's a test problem or that it is really necessary since they passed before.

image

PR #298, opened 2020-10-08
Add withConcurrentVar, modify and modify_

1Jajen1:
A few useful methods for manipulating ConcurrentVars safely using bracketCase.


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ConcurrentVar.kt
```diff
@@ -260,6 +293,33 @@ private class DefaultConcurrentVar<A> constructor(initial: State<A>) : Concurren
...
 
+  override suspend fun <B> withConcurrentVar(f: suspend (A) -> B): B =
+    bracketCase(
+      acquire = ::take,
+      use = f,
+      release = { a, _ -> put(a) }
+    )
+
+  override suspend fun <B> modify(f: suspend (A) -> Tuple2<A, B>): B {
+    // ugly. Is there a better way?
```
nomisRev, line 304: This is the fastest and simplest implementation afaik. Just a tad ugly yes 😅

nomisRev (approved):
Should we add some tests? It's `bracket` composition...

Could we add some docs to explain why these methods are favored like we discussed on Slack or in the STM PR?

1Jajen1:
Well it is just composition so I don't think being very exhaustive in the tests is necessary, although 1-2 basic tests can't hurt to at least show that it works.

I'll add something to docs, although this is best explained in the context of how to use `ConcurrentVar` in general. As in "Hey you want to use ConcurrentVar as a Lock? Here are some helpers that make sure you don't screw up with exceptions or cancellation".

nomisRev:
Great tests and docs, thanks @1Jajen1!

raulraja (approved):
Looks great, thanks Jannis,

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ConcurrentVar.kt
```diff
@@ -258,6 +300,33 @@ private class DefaultConcurrentVar<A> constructor(initial: State<A>) : Concurren
...
+      use = {
+        val (a, b) = f(it)
+        res = a
+        b
+      },
+      release = { a, exit ->
+        when (exit) {
+          is ExitCase.Failure -> put(a)
+          is ExitCase.Cancelled -> put(a)
+          is ExitCase.Completed -> put(res!!)
```
pakoito, line 324: `res?.let(::put)` maybe?

pakoito (approved):

danimontoya (approved):

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ConcurrentVar.kt
1Jajen1: Should `put(res!!)` ever fail for some reason I'd rather have it crash than not put back the value because that could lead to a very confusing deadlock.

PR #302, opened 2020-11-13
Add a lot more documentation to STM and structure it better

1Jajen1:
So after seeing the rendered api docs for STM I wasn't quite satisfied. It was really hard to see which methods worked on which datatype and what the datatypes actually did. 

This hopefully addresses that issue by documenting the datatypes with the methods used to operate on them. It also adds a ton of simple examples and changes/adds a few `operator` overload methods to be more like the stdlib collections.

Edit: This is 95-99% documentation btw^^ So don't be scared by `+2000` here :see_no_evil:


1Jajen1:
Oh and also: Is there a place in the `arrow-fx` docs sidebar that we can feature `STM` in? Apidocs is ok-ish but adding it to like datatypes as well would be a bit better imo. Just not sure it fits there.

nomisRev:
@1Jajen1 need to make some time to go through this, but feel free to restructure the Arrow Fx website a bit to make room for STM in the sidebar. Perhaps we need to relocate `data types` under an `Arrow Fx Coroutines` specific section?
We might also want to make room for more modules later on in a similar style.

https://github.com/arrow-kt/arrow-site/blob/master/docs/_data/sidebar-fx.yml

1Jajen1:
> We might also want to make room for more modules later on in a similar style.

Likely as this will apply to `Stream` as well.

I'll try to think of something, but as of now I don't have a good idea what the category for `STM` (or `Stream`) should be...

nomisRev:
We could change the `Data types` section to `Arrow Fx Coroutines`, and we can add a section for `STM` and `Stream`?
I am working on the split of Streams atm ;)

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMVar.kt
```diff
@@ -19,6 +19,184 @@ fun <A> STM.newEmptyTMVar(): TMVar<A> = TMVar<A>(newTVar(Option.None))
...
+ *     //sampleEnd
+ *     println("Result $result")
+ *     println("New value ${atomically { tmvar.tryTake() } }")
+ *   }
+ * }
+ * ```
+ *
+ * Another effect of using either [STM.take] or [STM.tryTake] is that the [TMVar] will be empty after a successful call. Alternatively if you just
+ *  want to read without emptying you can use either [STM.read] or [STM.tryRead] which will not remove the value.
+ *  As with [STM.take] [STM.read] will fail and retry the transaction should the [TMVar] be empty, whereas [STM.tryRead] will return null instead.
```
nomisRev, line 71:
```suggestion
 *  As with [STM.take] [STM.read] will fail and retry the transaction should the [TMVar] be empty, whereas [STM.tryTake] and [STM.tryRead] will return null instead.
```

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMap.kt
```diff
@@ -8,6 +8,171 @@ fun <K, V> STM.newTMap(fn: (K) -> Int): TMap<K, V> = TMap(newHamt(), fn)
...
+ *
+ *       tmap.lookup(1) + tmap[2]
+ *     }
+ *     //sampleEnd
+ *     println("Result $result")
+ *   }
+ * }
+ * ```
+ *
+ * > If the key is not present [STM.lookup] will not retry, instead it returns `null`.
```
nomisRev, line 51: I think we should add this to the definition of `lookup` and `operator fun get` in STM as well.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMVar.kt
```diff
@@ -19,6 +19,184 @@ fun <A> STM.newEmptyTMVar(): TMVar<A> = TMVar<A>(newTVar(Option.None))
...
+ *       tmvar.take()
+ *     }
+ *     //sampleEnd
+ *     println("Result $result")
+ *     println("New value ${atomically { tmvar.tryTake() } }")
+ *   }
+ * }
+ * ```
+ *
+ * Should the [TMVar] be empty at the time of calling [STM.take], it will call [STM.retry], suspend and wait for another transaction to
```
nomisRev, line 45: Should be added to the definition of `take` as well.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMVar.kt
```diff
@@ -19,6 +19,184 @@ fun <A> STM.newEmptyTMVar(): TMVar<A> = TMVar<A>(newTVar(Option.None))
...
+ *     val result = atomically {
+ *       tmvar.isEmpty()
+ *     }
+ *     //sampleEnd
+ *     println("Result $result")
+ *   }
+ * }
+ * ```
+ *
+ * > Because the state of a transaction is constant there can never be a race condition between checking if a `TMVar` is empty and subsequent
```
nomisRev, line 197: Should be added to the definition of `isEmpty` and `isNotEmpty` as well.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TQueue.kt
```diff
@@ -7,17 +7,244 @@ fun <A> STM.newTQueue(): TQueue<A> = TQueue(newTVar(PList.Nil), newTVar(PList.Ni
...
+ *     val result = atomically {
+ *       tq.size()
+ *     }
+ *     //sampleEnd
+ *     println("Result $result")
+ *   }
+ * }
+ * ```
+ *
+ * > All three of these methods have to access both the write and read end of a [TQueue] and thus can increase contention. Use them sparingly!
```
nomisRev, line 222: Should be added to the definition of these methods as well.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TQueue.kt
```diff
@@ -7,17 +7,244 @@ fun <A> STM.newTQueue(): TQueue<A> = TQueue(newTVar(PList.Nil), newTVar(PList.Ni
...
+ *       tq.filter { it != 0 }
+ *     }
+ *     //sampleEnd
+ *     println("Items in queue ${atomically { tq.flush() }}")
+ *   }
+ * }
+ * ```
  *
- * > In practice reads have to access the writes variable if they run out of elements to read but this is infrequent.
+ * > Both filtering methods also access both ends of the queue and thus should be used rarely to avoid contention.
```
nomisRev, line 247: Added to definition

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TSet.kt
```diff
@@ -8,6 +8,97 @@ fun <A> STM.newTSet(fn: (A) -> Int): TSet<A> = TSet(newHamt(), fn)
...
+ * ## Where are operations like `isEmpty` or `size`?
+ *
+ * This is a design tradeoff. It is entirely possible to track size however this usually requires one additional [TVar] for size and
+ *  almost every operation would modify that. That will lead to contention and thus decrease performance.
+ *
+ * Should this feature interest you and performance is not as important please open an issue. It is most certainly possible to add another version
+ *  of [TSet] that keeps track of its size.
+ *
+ */
+// Why a `Pair<Unit, A>`? Well because kotlin. Altering a Hamt is done through inline *reified* methods and because we cannot get
```
nomisRev, line 99: I'm not sure I understand why `Pair<Unit, A>` instead of just `A`, also not from looking at the inline *reified* methods. Also 😂

nomisRev (approved):
Great addition to the docs. Some suggestions, which leads to a bit to duplication but I think that's okay!

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TSet.kt
1Jajen1: Iirc `reified` needs some proof of a concrete class to compile otherwise the calling methods must be reified and thus inline as well. That is not possible on interfaces because.... well idk. Either way `Pair<Unit, A>` seems to fulfill this need for a concrete class and thus avoids having to make the interface methods inline.

Maybe there are better solutions, maybe even inline classes to get rid of the wrapper after the fact...

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
```diff
@@ -208,22 +209,90 @@ interface STM {
    * Rerun the current transaction.
    *
    * Aborts the transaction and suspends until any of the accessed [TVar]'s changed, after which the transaction restarts.
```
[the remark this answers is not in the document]
1Jajen1: When `retry` is called it first validates the transaction to see if it can restart right now. If not it registers itself at any accessed `TVar` (checking for changes before and after registering to not miss updates) and then it suspends until its `Continuation` is called from any changed `TVar`.

Also `TVar` is the only primitive datastructure, all others are built upon it.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
```diff
@@ -208,22 +209,90 @@ interface STM {
...
+   * suspend fun main() {
+   *   Environment().unsafeRunSync {
+   *     //sampleStart
+   *     val result = atomically {
+   *       stm { retry() } orElse { "Alternative" }
+   *     }
+   *     //sampleEnd
+   *     println("Result $result")
+   *   }
+   * }
```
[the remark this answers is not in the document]
1Jajen1: None of these examples ever needs more than the `suspend` keyword since they never actually suspend. (Except the ones for the general `STM` docs).

What happens if a suspend function run by `suspend fun main` suspends? Is there any coroutine machinery that would allow other coroutines to run and unblock that one? If not then I'd keep `Environment` so that anyone who copies this in can't shoot themselves in the foot while experimenting. But all of this is guesswork so :see_no_evil:

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMVar.kt
1Jajen1: It already is, though with different wording.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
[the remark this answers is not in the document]
nomisRev: Kotlin lowers `suspend fun main` into a regular `suspend` function by wrapping in `runSuspend` which on the JVM  spins until the `Contiuation` completes. So it's similar to `Enviroment().unsafeRunSync`, but I don't think we should promote the use of `unsafeRunSync` anywhere since it relies on a lock.

https://github.com/JetBrains/kotlin/blob/9adc1a6e9bd1f79ecc58988a659d093cf40f6434/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt#L75

https://github.com/JetBrains/kotlin/blob/deb416484c5128a6f4bc76c39a3d9878b38cec8c/libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/RunSuspend.kt

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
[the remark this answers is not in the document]
nomisRev: > Also TVar is the only primitive datastructure, all others are built upon it.
Yes, I was just wondering if we should repeat that here. Something like:

> Aborts the transaction and suspends until any of the accessed [TVar]'s changed, so all data structures found in the STM package , after which the transaction restarts.

I think the explanation you shared here on `retry` also more clear IMO!

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TSet.kt
nomisRev: 🤔 Not sure I understand what the difference is between `Hamt<A>` and `Hamt<Pair<Unit, A>>`, since it's wrapped in another class?

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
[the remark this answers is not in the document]
1Jajen1: So what should we do here? Just `suspend fun` to keep it simple or use one of our runners? Also I think `STM` has no actual dependency on `arrow-fx-coroutines` atm except for docs. Maybe I should remove that entirely so we can drop the dependency entirely...

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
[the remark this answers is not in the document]
1Jajen1: >> Also TVar is the only primitive datastructure, all others are built upon it.

> Yes, I was just wondering if we should repeat that here. Something like:

> > Aborts the transaction and suspends until any of the accessed [TVar]'s changed, so all data structures found in the STM package , after which the transaction restarts.

This is pretty similar to how GHC documents it, but I can see this being confusing and your addition does clear that up :+1: 

> I think the explanation you shared here on retry also more clear IMO!

It is, but imo it goes too much into implementation detail which I'd like to avoid. Not sure how to phrase it better :/

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TSet.kt
1Jajen1: Ah this is not about `Hamt<A>` not working, it is about `alterHamtWithHash` using methods like `arrayOf` which are `inline` and have `reified` generics. Because the elements used here are generic this requires `alterHamtWithHash` to be `inline` and use `reified` as well. The methods on `TSet` are implemented on the `STM` interface and use `alterHamtWithHash` and with a generic `A` as well. This means they would also need to be `inline` and use `reififed` but since we are on an interface that does not work. The solution is to supply `alterHamtWithHash` with a concrete class for the generic (`Pair` in this case) so that we don't need to propagate `inline` to the methods.

I should look into using an inline class here tho but iirc an inline class over generics may not disappear after compilation.

This sucks.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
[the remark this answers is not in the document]
nomisRev: If we can drop the dependency, then let's do that! IMO we should use `suspend fun` in the examples, to be more clear. Ideally, if cancellation ever drops in the standard library, then the need to use `Environment` should drop to almost 0%. At that point, it should be used exclusively to interop with Java APIs that do not support any other effect types. Even `CompletableFuture` is a better option than `Environment.unsafeRunSync`.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TSet.kt
nomisRev: Oh, I see now.. Thanks for the clarification! Not sure if `inline class` would help there, since I still expect it to do boxing.
The only other option I see is to use casting inside of `alterHamtWithHash` to remove the `reified` restriction on it.

use `arrayOf<Any?>(newEl, *branch.value) as Array<A>`, instead of `arrayOf<A>(newEl, *branch.value)` and you'll be able to get rid of `reified`. Not a must, and definitely not in this PR. It's a trick I used in `Chunk` in a couple of places as well to avoid `reified`.

1Jajen1: Ah that is a good idea! Will note that somewhere so I remember^^

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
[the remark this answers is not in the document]
1Jajen1: I changed all examples except the first few because they actually use `arrow-fx-coroutines`. I also changed the dep to just core and since ank is global and includes everything anyway the few examples that need `fx-coroutines` should work just fine I think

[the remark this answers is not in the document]
nomisRev: Awesome, just reviewed the new changes, and they look great 👏 👏 👏

nomisRev:
A second review is needed @arrow-kt/maintainers

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMap.kt
```diff
@@ -8,6 +8,153 @@ fun <K, V> STM.newTMap(fn: (K) -> Int): TMap<K, V> = TMap(newHamt(), fn)
...
 fun <K, V> STM.newTMap(hash: Hash<K>): TMap<K, V> = newTMap { hash.run { it.hash() } }
 
+/**
+ * A [TMap] is a concurrent transactional implementation of a key value hashmap.
+ *
+ * Based on a Hash-Array-Mapped-Trie implementation. While this does mean that a read may take up to 5 steps to be resolved (depending on how
+ *  well distributed the hash function is), it also means that structural changes can be isolated and thus do not increase contention with
+ *  other transactions. This effectively means concurrent access to different values is unlikely to interfere with each other.
+ *
+ * > Hash conflicts are resolved by chaining.
```
raulraja, line 18: What does it mean by chaining?

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMap.kt
```diff
@@ -8,6 +8,153 @@ fun <K, V> STM.newTMap(fn: (K) -> Int): TMap<K, V> = TMap(newHamt(), fn)
...
+ * ```kotlin:ank:playground
+ * import arrow.fx.stm.TMap
+ * import arrow.core.toT
+ * import arrow.fx.stm.atomically
+ *
+ * suspend fun main() {
+ *   //sampleStart
+ *   val tmap = TMap.new<Int, String>()
+ *   atomically {
+ *     tmap + (1 to "Hello")
```
raulraja, line 80: Should these be += instead of just plus if it returns Unit?

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TQueue.kt
```diff
@@ -7,17 +7,214 @@ fun <A> STM.newTQueue(): TQueue<A> = TQueue(newTVar(PList.Nil), newTVar(PList.Ni
...
+ * import arrow.fx.stm.atomically
  *
- * [TQueue] is implemented using two [TVar]'s. One for reading and one for writing.
- * This effectively means that writing and reading accesses two different variables and thus it never blocks each other.
+ * suspend fun main() {
+ *   //sampleStart
+ *   val tq = TQueue.new<Int>()
+ *   atomically {
+ *     tq.write(0)
+ *     tq.filter { it != 0 }
```
raulraja, line 210: filter as a name is confusing if it does not return a new queue as users may be used ro. Is there a better name for this op?

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TSem.kt
```diff
@@ -1,20 +1,98 @@
...
  *
  * ## Creating a [TSem]
  *
  * A [TSem] is created by using either [TSem.new] outside of transactions or [STM.newTSem] inside a transaction.
  * Both of these methods throw if the supplied initial value is negative.
+ *
+ * ## Acquiring one or more permits
+ *
+ * ```kotlin:ank:playground
+ * import arrow.fx.stm.TSem
```
raulraja, line 18: Can these be named TSemaphore or similar instead of the abbreviated names? I think for var is well known but the rest are not descriptive and can be confusing to users.

raulraja (approved):
This is great, thanks so much @1Jajen1 and @nomisRev for taking the time and reviewing. I requested some changes regarding type names and operator names to make it closer to our efforts to be inline with the std lib semantics and make some types like TSem more exploit if possible.

1Jajen1:
> I requested some changes regarding type names and operator names to make it closer to our efforts to be inline with the std lib semantics and make some types like TSem more exploit if possible.

Yes please, now is the perfect time to align the api, I took the names from the haskell libs and related so any change is fine by me. I'll do it in this pr if that is fine with you.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMap.kt
1Jajen1: There are two ways to to resolve hash conflicts:
- Either replace the value and assume hashes are unique
- Or put them in a collection and iterate through them on lookups

Chaining here refers to chaining it to the end of the leaf array, I've heard and seen it used in this context while I did a bit of research on how Hamt structures work so I just used it here.

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMap.kt
1Jajen1: That would actually make more sense :+1:

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TQueue.kt
1Jajen1: I honestly have no idea, it really does filter, so dunno. Actually I really dislike the filter as an api because it a) increases contention (accesses both TVars) and b) its not pure and has a weird/different api as you mentioned.

Maybe it makes sense to just :fire: it.

1Jajen1: Maybe `applyFilter` to highlight that it directly applies to the `TQueue`?

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TMap.kt
1Jajen1: Changed it, it does look a lot better^^

arrow-fx-stm/src/main/kotlin/arrow/fx/stm/TQueue.kt
nomisRev: If you dislike the function, I am fine with removing it. It can always be re-added later, removing is harder 😁
I agree the name is a bit confusing.
Perhaps something along the line of `remove`/`removeAll`? That's a function found in the Kotlin Standard library but only available for mutable collections.

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/remove.html
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/remove-all.html

1Jajen1: I think keeping it is fine, `removeAll` with a predicate is a really good name. It still has the problem of contention but it may have some actual use, so long as its used infrequently its perfectly fine.

nomisRev (approved):
👌

PhBastiani:
I'm starting to read the doc ... I'm impressed with the work you have done for Continuations & STM 👏

A README explaining the main functionalities would be useful to introduce the purpose of the module.

1Jajen1:
> I'm starting to read the doc ... I'm impressed with the work you have done for Continuations & STM 
> 
> A README explaining the main functionalities would be useful to introduce the purpose of the module.

Regarding the implementation or the api function? I suppose a readme describing how the implementation works ala `fx-coroutines` makes most sense as the rest should be in the docs^^

1Jajen1:
>> I'm starting to read the doc ... I'm impressed with the work you have done for Continuations & STM
A README explaining the main functionalities would be useful to introduce the purpose of the module.

> Thanks for reading it pray

> The README: Regarding the implementation or the api function? I suppose a readme describing how the implementation works ala fx-coroutines makes most sense as the rest should be in the docs^^

Either way I think its best left for another pr as this one is quite close to completion. For a general introduction to what STM can do the api docs for the `STM` interface and the documentation for `atomically` are the best start.

PhBastiani:
> Either way I think its best left for another pr as this one is quite close to completion. For a general introduction to what STM can do the api docs for the `STM` interface and the documentation for `atomically` are the best start.
We don't need a long README: a short introduction would be sufficient, as long as the reader is encouraged to read the STM class doc (I mean it is important to mark the main entry point of the API).

1Jajen1:
> We don't need a long README: a short introduction would be sufficient, as long as the reader is encouraged to read the STM class doc (I mean it is important to mark the main entry point of the API).

Ah ok, I was confused by what exactly you were referring to. I'll add that in this pr, thanks for the suggestion! At some point I'll add a section to it which explains the internals as a sort of contribution guide like `arrow-fx-coroutines` does atm, but thats a later pr.

nomisRev:
> At some point I'll add a section to it which explains the internals as a sort of contribution guide like arrow-fx-coroutines does atm, but thats a later pr.
❤️

1Jajen1:
Added a readme which contains a link to the `STM` interface api docs. We still need a section in the sidebar, but until I have a decent idea where, that can wait.

If there is nothing else I'll merge this pr later today :tada:

PR #303, opened 2020-11-13
Split stream into separate module

nomisRev:
As discussed in #270 we're also splitting of Streams into a seperate module, and we're moving common test utilities to a common test module.


arrow-fx-coroutines-test/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+
+apply plugin: 'kotlinx-atomicfu'
+
+apply from: "$SUB_PROJECT"
+apply from: "$DOC_CREATION"
+
+dependencies {
+    api "io.arrow-kt:arrow-core:$VERSION_NAME"
+    api project(":arrow-fx-coroutines")
+    api project(":arrow-fx-coroutines-stream")
```
1Jajen1, line 14: Should this not include `arrow-core-test`? Some of the generators defined below are included there iirc

arrow-fx-coroutines-stream/src/main/kotlin/arrow/fx/coroutines/stream/predef.kt
```diff
@@ -35,3 +35,8 @@ internal fun <A, B> Iterator<A>.flatMap(f: (A) -> Iterator<B>): Iterator<B> =
     override fun next(): B =
       (if (hasNext()) cur else empty).next()
   }
+
+/** Represents a unique identifier using object equality. */
+internal class Token {
+  override fun toString(): String = "Token(${Integer.toHexString(hashCode())})"
+}
```
1Jajen1, line 42: How unique are `hashCode()` calls to this class? I actually never tested that :sweat_smile:

1Jajen1 (approved):
Looks good to me :+1:

arrow-fx-coroutines-stream/src/main/kotlin/arrow/fx/coroutines/stream/predef.kt
nomisRev: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()

> As much as is reasonably practical, the hashCode method defined by class Object does return distinct integers for distinct objects. (This is typically implemented by converting the internal address of the object into an integer, but this implementation technique is not required by the JavaTM programming language.)

So "As much as is reasonably practical" 😅

arrow-fx-coroutines-test/build.gradle
nomisRev: `arrow-core-test` and co are still behind on KotlinTest :(

arrow-fx-coroutines-stream/src/main/kotlin/arrow/fx/coroutines/stream/Chunk.kt
```diff
@@ -973,9 +973,9 @@ abstract class Chunk<out O> {
               val next = rem.first()
               val nextSize = next.size()
               when {
-                nextSize < toDrop -> loop(rem.tail(), toDrop - nextSize)
-                nextSize == toDrop -> Queue(rem.tail(), size - n)
-                else -> Queue(next.drop(toDrop) prependTo rem.tail(), size - n)
+                nextSize < toDrop -> loop(rem.tailOrNull()!!, toDrop - nextSize)
```
aballano, line 976: I think the exception thrown by `tail` would be more descriptive than a NPE here, no? Otherwise I believe you could remove the `tail` function as it's not used anymore

arrow-fx-coroutines-stream/src/main/kotlin/arrow/fx/coroutines/stream/callbacks.kt
```diff
@@ -2,10 +2,10 @@ package arrow.fx.coroutines.stream
 
 import arrow.core.Either
 import arrow.fx.coroutines.CancelToken
+import arrow.fx.coroutines.UnsafePromise
```
aballano, line 5: nit: Not sure if this was done automatically, but it was right before

arrow-fx-coroutines-test/build.gradle
aballano, line 14: This dependency shouldn't be needed AFAIK. I believe you could move the StreamSpec/Arb classes to arrow-coroutines-stream/test directory and get rid of this dependency + leave arrow-coroutines-test focused only in `arrow-coroutines`

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
```diff
@@ -12,7 +12,7 @@ import kotlin.coroutines.startCoroutine
 const val ArrowExceptionMessage =
   "Arrow-kt internal error. Please let us know and create a ticket at https://github.com/arrow-kt/arrow/issues/new/choose"
 
-internal open class ArrowInternalException(override val message: String = ArrowExceptionMessage) : RuntimeException(message)
+open class ArrowInternalException(override val message: String = ArrowExceptionMessage) : RuntimeException(message)
```
aballano, line 15: I'm not sure how to feel about a globally-available-open arrow internal exception 😕

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
```diff
@@ -25,13 +24,13 @@ internal fun <A> Iterable<A>.deleteFirst(f: (A) -> Boolean): Pair<A, List<A>>? {
   return go(this, emptyList())
 }
 
-internal fun <A> Iterable<A>.uncons(): Pair<A, List<A>>? =
+fun <A> Iterable<A>.uncons(): Pair<A, List<A>>? =
   firstOrNull()?.let { Pair(it, drop(1)) }
 
-internal fun Iterable<*>.isEmpty(): Boolean =
+fun Iterable<*>.isEmpty(): Boolean =
```
aballano, line 30: Same here, not sure if it's a good idea to publish this extension, specially given the fact it gives no clue about it's doing a fold over the entire data set in case is not a Collection

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
```diff
@@ -25,13 +24,13 @@ internal fun <A> Iterable<A>.deleteFirst(f: (A) -> Boolean): Pair<A, List<A>>? {
   return go(this, emptyList())
 }
 
-internal fun <A> Iterable<A>.uncons(): Pair<A, List<A>>? =
+fun <A> Iterable<A>.uncons(): Pair<A, List<A>>? =
```
aballano, line 27: this function and the ones above should probably be moved to arrow-syntax?

arrow-fx-coroutines-stream/src/main/kotlin/arrow/fx/coroutines/stream/Chunk.kt
nomisRev: `tail` is indeed not used anymore, just had to double check. We could move it as an extension method to the `Stream` module instead and keep the more descriptive error message.

arrow-fx-coroutines-test/build.gradle
nomisRev: Yes, that would be possible. I included this since not all the custom `Arb` instances are easy to write for clients, and I didn't want to include even more modules 🙈

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
nomisRev: It's used across several modules in Arrow Fx Coroutines, and I didn't see a big problem which such a type being visible for users. I'm not sure which issues this could cause?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
nomisRev: Just double-checked, and this could be moved to Stream as `internal`. I did however really miss an entire set of these methods in the Kotlin Std given that you can find them in Scala Std.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
nomisRev: IMO, there is an entire set of `Iterable` operation missing from Kotlin Std or Arrow Syntax. This could also be isolated to Streams.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
aballano: From an api design perspective the users don't need to be able to instantiate or extend an internal exception. This is normally fixed by having a common internal module with all our internal exceptions, functions, etc. that are only exposed to arrow itself but not outside. Although I understand it's out of the scope of this PR I believe we should discuss (soon) which way we want to go in a ticket/thread

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
nomisRev: I didn't see a big issue here, since these are also all things we find Scala Std, Guava, Apache Common, etc
IMO there is a huge set of these `Iterable` operators missing from Kotlin's Std. We could duplicate `size`, and move some of the other methods to Stream.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
aballano: Maybe is up for a discussion/ticket, my choices are: make them internal for our use or provide them with proper docs and all in arrow-syntax

nomisRev: Moved to streams

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/predef.kt
nomisRev: Moved to streams as internal and duplicated `size`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
nomisRev: Duplicated and made internal.

arrow-fx-coroutines-stream/src/main/kotlin/arrow/fx/coroutines/stream/Chunk.kt
nomisRev: Moved `tail` to Streams, and gave a more descriptive message.

arrow-fx-coroutines-stream/src/main/kotlin/arrow/fx/coroutines/stream/predef.kt
```diff
@@ -0,0 +1,82 @@
...
+      }
+      return true
+    }
+
+    override fun next(): B =
+      (if (hasNext()) cur else empty).next()
+  }
+
+internal const val ArrowExceptionMessage =
+  "Arrow-kt internal error. Please let us know and create a ticket at https://github.com/arrow-kt/arrow/issues/new/choose"
```
aballano, line 42: image

aballano (approved):

rachelcarmena:
Thanks @nomisRev :raised_hands: The build doesn't finish in 90 min because of:

```
arrow.fx.coroutines.stream.ParJoinTest > outer failed FAILED
    TimeoutException(duration=600000)
```

Could it be removed or tuned?

PR #307, opened 2020-11-27
Remove running on `EmptyCoroutineContext

nomisRev:
Running tokens on `EmptyCoroutineContext` removes the `CoroutineContext` from the `suspend` program, and potentially makes us lose state. Instead we should always favor `Unintercepted` + `coroutineContext`/`Continuation#context` to run `suspend` code on current thread with existing `CoroutineContext` state.


1Jajen1 (approved):
This looks good to me, its a step forward to not loosing the context. This may have some unforeseen effects but so did the use of `EmptyCoroutineContext` and this is definitely better^^

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ForwardCancellable.kt
```diff
@@ -18,35 +18,37 @@ import kotlin.coroutines.suspendCoroutine
...
  * in [uncancellable] to back-pressure the original context until the uncancellable context is finished.
+ *
+ * @param ctx, the CoroutineContext where the `CancelToken` is run on.
+ *   => This is taken into the constructor because we want to run it on the `CoroutineContext` of the creator of the `CancelToken`,
+ *   and not on the `CoroutineContext` of the caller of `cancel`.
+ *   => We want the `CancelToken` from `cancellable` to run on the `CoroutineContext` where `cancellable` is  called.
+ *   => We want the release function of `bracket` & `bracketCase` to run on the `CoroutineContext` where `bracketX` is called.
  */
-internal class ForwardCancellable {
+internal class ForwardCancellable(private val ctx: CoroutineContext) {
```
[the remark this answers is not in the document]
nomisRev: Yes, this could definitely be configurable. I'm still wondering a bit, since running cancel tokens on an `IOPool` is probably safest.

Now the're most likely to run on the default pool, but for the current usages I think this makes most sense. It's used in `bracketCase` and it's derived methods, and `cancellable`. It makes the `release` and `CancelToken` to be run with the same `ctx` as their `use/acquire` and `f` functions.

danimontoya (approved):
Neat 💯 !

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
```diff
@@ -159,6 +159,11 @@ object Platform {
     }
   }
 
+  fun composeErrors(first: Throwable, res: Result<Any?>): Throwable {
+    res.fold({ first }, { e -> first.addSuppressed(e) })
+    return first
+  }
```
aballano, line 165:
```suggestion
  fun composeErrors(first: Throwable, res: Result<Any?>): Throwable =
    res.fold({ first }, { e -> first.addSuppressed(e); first })
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/timer.kt
```diff
@@ -80,25 +77,30 @@ suspend fun <A> timeOutOrNull(duration: Duration, fa: suspend () -> A): A? =
     // Register our new tokens to our parents connection
     conn.push(listOf(suspend { timerConn.cancel() }, suspend { faConn.cancel() }))
 
-    suspend { sleep(duration) }.startCoroutineCancellable(
-      CancellableContinuation(
-        EmptyCoroutineContext,
-        timerConn
-      ) { timeOut ->
+    // Launch timer on current thread (Unintercepted) with default ctx (sleep returns there), and timer connection
+    // Launch on current thread, since it will immediately fork to sleeper scheduler, and free current Thread
```
aballano, line 81: I don't get this part, aren't you launching below on the Computation pool thread and not the current one?

aballano (approved):

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/timer.kt
nomisRev: `suspend { sleep(duration) }` is started with `startCoroutineUnintercepted` which means it immediately starts running without getting `intercepted()` which means it won't get scheduled on` ComputationPool`.

However `sleep` requires a `ContinuationInterceptor` to be installed, otherwise, it cannot shift away from the sleep scheduler when it returns from sleeping. This is problematic since nothing is supposed to run on the scheduling sleep.

So by using `startCoroutineUnintercepted` and `ComputationPool` we can launch the sleeping fiber without first intercepting/scheduling, and we install `ComputationPool` as a `ContinuationInterceptor` for `sleep` to safely return there.

PR #309, opened 2020-11-30
Add an option to switch out the default context used in arrow

### LINKED ISSUE #306: ["Request"] Make the default coroutine context configurable
Currently a whole bunch of methods use the `ComputationPool` context to execute tasks, and in some cases even do so without allowing a user to change it (which I would consider a bug).

There are a few use cases that would benefit from this added option:
- Interop with other libraries. Especially libraries that offer concurrency primitives and rely on some sort of contextual state
- Testing. With full control over the thread pools that arrow uses we can do ungodly things to prove code is correct. Also the whole advance time thing from kotlinx-coroutines-test gets somewhat easy to implement.
- Performance and resource tuning: Our computation pool may not offer the best performance for all applications and those that want to tune it should be able to!

A list of our methods that currently don't supply contexts and offer no defaults so they always end up using the `ComputationPool`:
 * Stream.Companion.effect (through raceN)
 * Stream.Companion.cancellable (through ForkAndForget)
 * Stream.interruptWhen (through ForkAndForget)
 * ForkScoped (through ForkConnected)
 * Scope.InterruptContext.childContext (through ForkConnected twice)
 * SignallingAtomic (almost every method through ForkConnected)

These methods aren't the only ones that would need changes, these are just the ones that don't respect or don't give the user options to not use the `ComputationPool` which imo is not desirable. 

I also talked about this with @nomisRev and he had the idea to add the current default pool to the coroutine context and retrieve it from there. That may be the simplest and easiest implementation for this.

1Jajen1:
Currently we use `ComputationPool` everywhere and its mostly hardcoded. This enables us to switch the default context out for a different one at the edge of our program using `Environment(ctx)`.

This is **not** yet done: The code in here is hopefully done, but it also needs to be reflected in the docs. I am just putting this out there to get some feedback on this solution :see_no_evil: 

Closes #306


nomisRev:
This is a great follow up for #307, there is an additional `ComputationPool` usage there that needs to be replaced in `timer.kt`. 😁

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Race2.kt
```diff
@@ -109,7 +109,8 @@ suspend fun <A, B> raceN(ctx: CoroutineContext, fa: suspend () -> A, fb: suspend
     err: Throwable
   ): Unit =
     if (active.getAndSet(false)) {
-      suspend { other.cancel() }.startCoroutine(Continuation(ComputationPool) { r2: Result<Unit> ->
+      // Why?
+      suspend { other.cancel() }.startCoroutine(Continuation(EmptyCoroutineContext) { r2: Result<Unit> ->
```
nomisRev, line 113: Fixed in #307

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -98,3 +100,15 @@ internal class DefaultEnvironment(override val ctx: CoroutineContext) : Environm
...
 }
+
+private fun CoroutineContext.asDefaultCtx(): CoroutineContext = plus(DefaultContext(this))
+
+internal class DefaultContext(val ctx: CoroutineContext) : AbstractCoroutineContextElement(DefaultContext) {
+  companion object Key : CoroutineContext.Key<DefaultContext>
+}
+
+internal inline fun CoroutineContext.defaultContext(ctx: CoroutineContext): CoroutineContext =
+  this[DefaultContext] ?: ctx
```
nomisRev, line 111:
```suggestion
  this[DefaultContext]?.ctx ?: ctx
```

Need to unwrap from `DefaultContext`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
```diff
@@ -98,3 +100,15 @@ internal class DefaultEnvironment(override val ctx: CoroutineContext) : Environm
...
+
+internal suspend inline fun getDefaultContext(ctx: CoroutineContext = ComputationPool): CoroutineContext =
```
nomisRev, line 113: Currently, this param is always passed and it's always `ComputationPool`. Should we hardcode it and remove the param?

nomisRev (approved):
Looks good to me, should we add a note about `DefaultContext` in the doc of `Enviroment`?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
1Jajen1: I had this fixed in my test context branch and completely forgot to put this here, thanks! :see_no_evil:

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Enviroment.kt
1Jajen1: Yeah that could clean this up a bit^^

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/CancellableF.kt
```diff
@@ -28,7 +28,7 @@ class CancellableF : ArrowFxSpec(spec = {
         latch.complete(Result.success(Unit))
         CancelToken { cancelled.complete(true) }
       }
-    }.startCoroutineCancellable(CancellableContinuation { })
+    }.startCoroutineCancellable(CancellableContinuation(coroutineContext) { })
```
nomisRev, line 31: This deadlocks since [`coroutineContext` is a single threaded context](https://github.com/kotest/kotest/blob/cf7a5ed39361cef281d6494205091db7fc7ff491/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/KotestEngine.kt#L132) given the default configuration from Kotest is = 1, and `c.invoke()` is performing `unsafeRunSync` (with lock) to perform a back-pressured cancellation.

This can be fixed by `ForkConnected { c.invoke() }` but we should perhaps also evaluate if we want to keep this cancellation model. We should perhaps evaluate if that is the only way we want to expose cancellation, it would probably be safer to expose both `suspend fun cancel(): Unit` and `fun unsafeCancel(): Unit` where `unsafeCancel(): Unit` doesn't use `unsafeRunSync` behind the scenes but rather `unsafeRunAsync` which would remove the requirement of a lock and would not make this deadlock.

nomisRev: This would also fixes it, which keeps the same semantics but suspends instead of locks.

```kotlin
val cancelled = Promise<Boolean>()
val latch = UnsafePromise<Unit>()
val conn = SuspendConnection()
val c = suspend {
  cancellable<Unit> {
    latch.complete(Result.success(Unit))
    CancelToken { cancelled.complete(true) }
  }
}.startCoroutineCancellable(CancellableContinuation(coroutineContext, conn) { })
latch.join()
conn.cancel()
cancelled.get() shouldBe true
```

1Jajen1: Hmm I have `cancelableF executes generated task uninterruptedly` failing locally as well. I don't think I understand the tests well enough to fix them just yet. I'll look at this when I have some more time later this week^^

nomisRev:
@1Jajen1 this has become obsolute with the PR that introduces KotlinX Coroutines as a base dependency. However KotlinX offers testing support: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/.

nomisRev:
Closing because this since it became obselete by #317. Thanks @1Jajen1!

PR #310, opened 2020-12-01
Fix flaky test with empty stream

nomisRev:
We test that if the `concurrently` launched fiber fails, it also fails the parent `Stream`.
When the parent `Stream` finishes it also cancels the `concurrently` launched `Stream`.

In the case that the parent `Stream` is empty, it's possible that the fiber gets canceled before it got a chance to be scheduled. This makes the test fail since it results in `Unit` instead of a failure.


raulraja (approved):

aballano (approved):

rachelcarmena:
Thanks Simon!! :raised_hands:

PR #317, opened 2020-12-10
Add KotlinX Coroutines as base dependency to Arrow Fx Coroutines

nomisRev:
Currently in the eco-system we see that KotlinX Coroutines is hardcoded everywhere in frameworks. Spring, Ktor, Android, SqlDelight, etc and thus in all places where we care or want to grow. Since suspend is very implicit, in the sense that you can bind directly in the environment it’s important you mix the environments correctly. That’s something we currently cannot do without depending on KotlinX Coroutines, so we do so through a separate artifact and the responsibility lies with the user to do it correctly. This inevitably leads to a lot of strange bugs, which we’ve already seen on Kotlin Slack.

This can all be solved by depending on KotlinX Coroutines since we then can guarantee that the effect enviroments are correctly mixed, and thus no such strange bugs can occur in the user code.

Additionally, the maintenance and future development costs is drastically reduced since we don’t have to build or maintain interopt modules such as JDK8, Reactive Streams, RxJava, Reactor, etc since those live in the KotlinX Coroutines space. And in some areas such as generated code it’s even impossible for us to build interopt such as Spring which relies on generated code, or SqlDelight and perhaps Compose.

As mentioned above all the frameworks we care about already relied on KotlinX Coroutines so this is on everyones classpath anyway already.

This makes Arrow Fx Coroutines the companion library to KotlinX Coroutines for functional operators, such as Arrow Core is a companion to Kotlin’s standard library.

TODO:
 - Flakyness in `concurrently` Stream operator


arrow-benchmarks-fx/build.gradle
```diff
@@ -100,3 +100,7 @@ compareBenchmarksCI {
 repositories {
     maven { url "https://jitpack.io" }
 }
+
+compileKotlin {
+    kotlinOptions.jvmTarget = "1.8"
+}
```
rachelcarmena, line 106: Thanks @nomisRev !! It shouldn't be necessary to include this conf because it should come from https://github.com/arrow-kt/arrow/blob/master/gradle/subproject.gradle that is applied via `apply from: "$SUB_PROJECT"` Please, could you try it without it?

nomisRev: Yes, I think this is fine. It keeps showing up in my IDEA as an error, and sometimes will refuse to build from IDEA while it'll build fine from CLI.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
```diff
@@ -1,163 +1,18 @@
...
-
-    latch.acquireSharedInterruptibly(1)
-
-    return when (val either = ref) {
-      is Either.Left -> throw either.a
-      is Either.Right -> either.b
-      null -> throw ArrowInternalException("$ArrowExceptionMessage\nSuspend execution should yield a valid result")
-    }
-  }
+  internal fun <A> unsafeRunSync(f: suspend () -> A): A =
```
raulraja, line 14: Should we also deprecate this and just let people use runBlocking as unsafe method if they wish too? I think most of the users know this by now/

raulraja (approved):
Great improvement @nomisRev Kudos. This gives us a better maintenance path forward and much easier for users where it all works drop-in with KotlinX 👏

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ConcurrentVar.kt
```diff
@@ -42,6 +42,8 @@ import kotlinx.atomicfu.atomic
  * > Note that this only works if all operations over the [ConcurrentVar] follow the pattern of first taking and then putting back both exactly once and in order.
  *  Or use the helpers to also be safe in case of exceptions and cancellation.
  */
+// TODO benchmark against Channel(1), if Channel(1) is faster deprecate ConcurrentVar.
+// Channel(1) uses locks.
 interface ConcurrentVar<A> {
```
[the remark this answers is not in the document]
nomisRev: `ConcurrentVar` is also the same as a bounded `Queue` of size. So this type was already slightly redundant 😅

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Platform.kt
nomisRev: Yes, this has always been internal but that's why I `@Deprecated` `Enviroment` type since you should just use structured concurrency to run unsafely side-effects from Java or `runBlocking` to run from `main` or similar.

arrow-fx-coroutines-stream/src/test/kotlin/arrow/fx/coroutines/stream/CallbackTest.kt
```diff
@@ -206,17 +203,17 @@ class CallbackTest : StreamSpec(iterations = 250, spec = {
...
-private fun <A> countToCallback(
+private suspend fun <A> countToCallback(
   iterations: Int,
   map: (Int) -> A,
   cb: suspend (A) -> Unit,
   onEnd: suspend () -> Unit = { }
-): Unit = suspend {
+): Unit {
   var i = 0
   arrow.fx.coroutines.repeat(Schedule.recurs(iterations)) {
```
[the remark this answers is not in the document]
nomisRev: Yes, it could be a method on `Schedule` itself.

```
Schedule.recurs(iterations).repeat {
  ...
}
```

Only annoying is that in compositions you'd need to add parenthesis around it.

`(Schedule.recurs(iterations) zipRight Schedule.identity()).repeat { ... }` it might also worsen type inference since now `repeat` can infer which type `Schedule` has  to be based on the return type of lambda. Which is not possible in the  latter case, right? 🤔

rachelcarmena (approved):

raulraja:
Thanks @nomisRev !

PR #323, opened 2020-12-12
Fix computation block in parTraverse

nomisRev:
Fixes:

 > e: /arrow-fx/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ParTraverse.kt: (196, 3): Type mismatch: inferred type is Either<E, List<Unit>> but Either<E, List<B>> was expected
e: /arrow-fx/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ParTraverse.kt: (197, 5): Type mismatch: inferred type is List<Unit> but List<B> was expected
e: /arrow-fx/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/ParTraverse.kt: (198, 26): Unresolved reference: !

PR #332, opened 2020-12-14
inline deprecated BindSyntax members

i-walker:
following up with #323


nomisRev (approved):
These are not using computation blocks as mentioned in #323, these are using `Monad#fx` which are being deprecated in `0.12.0` and removed in `0.13.0`.

I'm fine merging these changes tho. Thanks @i-walker! 👏 👏

danimontoya (approved):
👌 As a suggestion we could also use the implicit way `()` without calling `invoke()`

PR #336, opened 2020-12-16
Reduce iterations property tests and stack safety iterations

nomisRev:
## Status
**READY**

## Description
Since `arrow-fx` (`IO` and effect typeclasses) are getting deprecated this version, and development has halted since last version we can safely reduce the iterations of the property based tests. This will reduce stress on CI, and time it takes to finish the CI actions. This also reduces the stress of Arrow Mtl which also runs effect typeclass tests.

Stack safety can be safely test with 5k iterations as well.

## Todos
- [x] Tests
~- [ ] Documentation~

## Related PRs
* None


i-walker (approved):
🙌  That makes sense !

raulraja (approved):
thanks!

1Jajen1 (approved):
Not sure if 5k is enough to find stackoverflows in some weirder scenarios, but tbh I would also be fine with just not running tests for `arrow-fx` at all once its fully deprecated and unlikely to be changed anymore.

rachelcarmena (approved):
Good idea!!! :raised_hands:

PR #338, opened 2020-12-16
Deprecate arrow.fx.coroutines.Duration

i-walker:
## Status
**On HOLD**

## Description

Deprecate and remove usages of arrow's Duration from all API's besides `arrow.fx.Timer` and other usages within `arrow.fx`, as it will be deprecated and removed soon. 

* Instead `kotlin.time.Duration` is used.
  In addition, this PR's centralizes some common build settings in
  - project("arrow-fx-coroutines"),
  - project("arrow-fx-coroutines-kotlinx-coroutines"),
  - project("arrow-fx-coroutines-test"),
  - project("arrow-fx-coroutines-stream"),
  - project("arrow-fx-stm")

## Todos
- [ ] Tests
- [ ] Documentation


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/timer.kt
```diff
@@ -15,7 +17,7 @@ import kotlinx.coroutines.withTimeoutOrNull
  * }
  * ```
  **/
-@Deprecated("Use delay", ReplaceWith("delay(duration.millis)", "kotlinx.coroutines.delay"))
+@Deprecated("Use delay", ReplaceWith("delay(duration.millis)", "kotlinx.coroutines.delay"), DeprecationLevel.ERROR)
```
nomisRev, line 20:
```suggestion
@Deprecated("Use delay", ReplaceWith("delay(duration.millis)", "kotlinx.coroutines.delay"))
```

Deprecating as `ERROR` breaks the users build, first round of deprecation is always `WARNING`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/timer.kt
```diff
@@ -37,6 +39,32 @@ suspend fun sleep(duration: Duration): Unit =
...
+ *     "Message from lazy task"
+ *   }.also(::println)
+ *
+ *   timeOutOrNull(2.seconds) {
+ *     "Message from fast task"
+ *   }.also(::println)
+ * }
+ * ```
+ **/
+suspend fun <A> timeOutOrNull(duration: kotlin.time.Duration, fa: suspend () -> A): A? =
```
nomisRev, line 69: Users should migrate to KotlinX `withTimeoutOrNull`.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/EvalOnTests.kt
```diff
@@ -36,7 +37,7 @@ class EvalOnTests : ArrowFxSpec(spec = {
 
   "evalOn on the same context doesn't dispatch" {
     suspend fun onSameContext(): String =
-      evalOn(ComputationPool) {
+      withContext(ComputationPool) {
```
nomisRev, line 40: This is just testing code from another library now.
If we want to make such a change we instead just remove `EvalOnTests` from `Arrow Fx Coroutines` since `evalOn` now only delegates to `withContext`.

nomisRev (changes requested):
The added function for `kotlin.time.Duration` in `timer.kt` should be removed. We don't want to add new code there since KotlinX already covers all of these APIs.

I don't understand the diff from ` arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/TimerTest.kt`. Is it re-added as an empty file?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Duration.kt
```diff
@@ -1,19 +1,35 @@
+@file:Suppress("DEPRECATION_ERROR")
+
 package arrow.fx.coroutines
 
 import java.util.concurrent.TimeUnit
 
-// TODO replace with kotlin.time
+@Deprecated(
+  "Redundant since 0.12.0 will provide automatic integration with KotlinX",
```
nomisRev, line 8: I think we need to provide a clearer message here. `Duration` is `Deprecated` in favor of `kotlin.time.Duration`, and KotlinX offers all APIs we offer in `timer.kt` for both `kotlin.time.Duration` and `milliseconds: Long` so APIs are available to the user with and without optining in for `kotlin.time`.

But that's a bit too long for a deprecated message I guess 😅

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/timer.kt
i-walker: I will add that for the others, too ! :D

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/timer.kt
i-walker: I will remove that :D

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/EvalOnTests.kt
i-walker: Ok I will do that

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Duration.kt
i-walker: Yes, I was also unsure, what message would make sense in this case. 
But I will go for yours, regardless 😄

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/evalOn.kt
```diff
@@ -5,7 +5,7 @@ import kotlin.coroutines.CoroutineContext
 
 @Deprecated("Use withContext", replaceWith = ReplaceWith("withContext(ctx, block)", "kotlinx.coroutines.withContext"))
```
nomisRev, line 6:
```suggestion
@Deprecated("Use withContext", replaceWith = ReplaceWith("withContext<T>(ctx) { this@evalOn.invoke() }", "kotlinx.coroutines.withContext"))
```

arrow/scripts/commons4gradle.sh
```diff
@@ -0,0 +1,190 @@
+#!/bin/bash
```
nomisRev, line 1: What is this file doing? Is this change needed? cc\\ @rachelcarmena

i-walker: No it is not needed, I wanted to execute the `buildArrowDoc` gradle task and it couldn't find that file from my local instance. I will remove that. 

Still I couldn't execute the gradle task.  
Do you know how I can perform the Ank checks?

nomisRev: There is a `buildArrowDoc` task available for me when checking out Arrow Fx 🤔 I think it relies on the other project being available locally, so `./gradlew gitPullAll`. Not 100% sure how that works. It's mostly easier to just run Ank on CI and fix all the issues that it raises 😅
That's why I made Ank work with `Validated` so long ago :joy:

i-walker:
This PR's is getting too big, so I am breaking it into multiple sub PR's to show some of the Error's I am running into

PR #357, opened 2020-12-22
Deprecates IO data type and related type classes in favor of Arrow Fx Coroutines

raulraja:
## Description
Deprecates IO data type and related type classes for 0.12.0

I adapted the current docs to remove all references to IO and its type classes. We still need to come up with better docs and structure for all the libs that would come in other PRs. 

## Todos
- [x] Documentation

## Related PRs
https://github.com/arrow-kt/arrow-core/pull/275

PR #358, opened 2020-12-23
inline sleep and timeOutOrNull

i-walker:
- [X] allow experimental time Api in build settings
- [X] Inline sleep/ timeOutOrNull
- [X] Replace library usage of arrow.fx.coroutines.Duration to kotlin.time.Duration, besides in API Types, such as CircuitBreak or Schedule

## Status
**Ready**

## Description
In order to break the changes introduced in #338 into multiple PR's
Api's that still use arrow's Duration 
- arrow.fx.coroutines.CircuitBreak addressed in #360 
- arrow.fx.coroutines.Schedule addressed here #359 

This PR does not change usages in arrow.fx, because it will be removed #357 

## Related PRs
* #338


nomisRev (approved):
Looks great. Thanks @i-walker!

1Jajen1 (approved):

PR #360, opened 2020-12-23
Replace arrow's Duration in CircuitBreaker to kotlin.time.Duration

i-walker:
## Status
**READY**

Related to #358


nomisRev (approved):
Perfect! 🎉

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
```diff
@@ -2,8 +2,11 @@ package arrow.fx.coroutines
 
 import arrow.core.Either
 import arrow.core.identity
+import kotlin.time.Duration
```
nomisRev, line 5: Breaking change

nomisRev (changes requested):
This breaks the APIs with arrow.fx.coroutines.Duration. These should be deprecated properly first. So we need to duplicate the signatures with arrow.fx.coroutines.Duration and deligate them to the new Kotlin Duration based impl.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
i-walker: Ok, I will create duplicated the descriptors with arrow.fx.Duration and create a PR for deprecating arrow.fx.coroutines.Duration

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
```diff
@@ -158,9 +161,9 @@ class CircuitBreaker constructor(
           }
           is ExitCase.Failure -> {
             // Failed reset, which means we go back in the Open state with new expiry val nextTimeout
-            val value = (resetTimeout.millis * exponentialBackoffFactor).toLong().milliseconds
-            val nextTimeout =
-              if (/*maxResetTimeout.isFinite &&*/ value.nanoseconds > maxResetTimeout.nanoseconds) maxResetTimeout
+            val value: Duration = (resetTimeout * exponentialBackoffFactor)
+            val nextTimeout: Duration =
+              if (/*maxResetTimeout.isFinite &&*/ value > maxResetTimeout) maxResetTimeout
```
nomisRev, line 166:
```suggestion
              if (maxResetTimeout.isFinite() && value > maxResetTimeout) maxResetTimeout
```

This is available on `kotlin.time.Duration` and is why I left it commented out ;)

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
```diff
@@ -5,6 +5,9 @@ import arrow.core.identity
 import arrow.fx.coroutines.CircuitBreaker.State.Closed
 import arrow.fx.coroutines.CircuitBreaker.State.HalfOpen
 import arrow.fx.coroutines.CircuitBreaker.State.Open
+import kotlin.time.Duration
+import kotlin.time.milliseconds
+import kotlin.time.nanoseconds
 
 class CircuitBreaker constructor(
```
nomisRev, line 12: The new constructor should be `private`, so people are forced to use the `of` constructor.
We should also deprecate the original constructor like we did for the `ADT`, but point to the `of` method with `ReplaceWith` instead of this new one.

```suggestion
class CircuitBreaker private constructor(
```

nomisRev (commented):
Thanks for these changes @i-walker 👏 👏
A couple nits, we should fix and this is ready to be merged!

nomisRev (approved):
Perfect, now there aren't any breaking changes 👌

danimontoya (approved):
Looks great! 🙌

PR #364, opened 2020-12-29
Improve Circuitbreak Api

### LINKED ISSUE #342: ["Request"] Improve CircuitBreaker API

0.11.0/0.12.0-SNAPSHOT

What would you like to see?

Deprecate `suspend fun <A> protect(f: suspend () -> A): A` and add with `protectOrThrow`, `protectOrNull` and `protectEither` instead. `protect` should include `ReplaceWith("protectOrThrow(fa)")` for smooth migrations.

i-walker:
## Issue
resolves #342 

## Status
**READY**


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
```diff
@@ -40,10 +41,26 @@ class CircuitBreaker constructor(
       is State.HalfOpen -> curr.awaitClose.get()
     }
 
-  /** Returns a new task that upon execution will execute the given
+  /**
+   * Returns a new task that upon execution will execute the given
    * task, but with the protection of this circuit breaker.
    */
-  tailrec suspend fun <A> protect(fa: suspend () -> A): A =
+  @Deprecated("#protect is deprecated in favor of protectOrThrow", ReplaceWith("protectOrThrow"))
```
nomisRev, line 48:
```suggestion
  @Deprecated("#protect is being renamed to protectOrThrow", ReplaceWith("protectOrThrow(fa)"))
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
```diff
@@ -40,10 +41,26 @@ class CircuitBreaker constructor(
...
+   * task, but with the protection of this circuit breaker.
+   * If an exception in [fa] occurs it will be rethrown
+   */
+  suspend fun <A> protectOrThrow(fa: suspend () -> A): A =
+    protectEither(fa).fold({ throw it }, ::identity)
+
+  suspend fun <A> protectOrNull(fa: suspend () -> A): A? =
+    protectEither(fa).fold({ null }, ::identity)
+
+  tailrec suspend fun <A> protectEither(fa: suspend () -> A): Either<Throwable, A> =
```
nomisRev, line 63: This should be `protectOrThrow`, and `protectEither` should wrap in `Either.catch`. That way we can avoid the allocations in the base implementation. `protectOrNull` can just `try/catch` and return `null`.

nomisRev (changes requested):

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
```diff
@@ -40,10 +41,36 @@ class CircuitBreaker constructor(
...
+  /**
+   * Returns a new task that upon execution will execute the given
    * task, but with the protection of this circuit breaker.
+   * If an exception in [fa] occurs it will be rethrown
    */
-  tailrec suspend fun <A> protect(fa: suspend () -> A): A =
+  suspend fun <A> protectOrNull(fa: suspend () -> A): A? =
+    try {
+      protectOrThrow(fa)
+    } catch (ex: Throwable) {
```
nomisRev, line 63:
```suggestion
    } catch (ex: ExecutionRejected) {
```

I'm wondering for `protectOrNull`, it should return `null` when `fa` was rejected or it should throw the original exceptions. For example for `fa = suspend { throw RuntimeException() }` we expect `RuntimeException` and not `null`.

cc\\ @raulraja @franciscodr

nomisRev (approved):
Looks great! Thanks @i-walker

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
raulraja: I think original exceptions should be thrown as `null` does not contain enough information as to what happened there and the `null` return can be a normal return of a user computation that returns null. Returning null in all cases won't allow to disambiguate whether a task finished exceptionally or not and we don't have here Either or anything similar to get a proper type info about exceptional cases.

raulraja (approved):

Issue #326, opened 2020-12-12
Flow retry operator using `Schedule`


nomisRev:
An operator that allows to `retry` a `Flow` for a given `Schedule`.

`fun <T, B> Flow<T>.retry(schedule: Schedule<Throwable, B>): Flow<T>`

PR #378, opened 2021-01-16
Flow.retry using Schedule

LordRaydenMK:
## Issue
Fixes: https://github.com/arrow-kt/arrow-fx/issues/326

## Status
**READY**

## Description
Retry operator for `Flow` using `Schedule`

## Todos
- [x] Tests
- [x] Documentation

Thanks @nomisRev for your help and support.


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/flow.kt
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @see [Schedule] for how to build a schedule.
+ *
+ * ```kotlin:ank:playground
+ * import kotlinx.coroutines.flow.*
+ * import arrow.fx.coroutines.*
+ * suspend fun main(): Unit {
+ *   val flow = flow {
+ *    emit(a)
+ *    if (++counter <= 5) throw RuntimeException("Bang!")
```
nomisRev, line 22: Where is this `counter` coming from?

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/FlowTest.kt
```diff
@@ -0,0 +1,66 @@
...
+        emit(a)
+        if (++counter <= 5) throw RuntimeException("Bang!")
+      }.retry(Schedule.recurs(n))
+        .reduce { acc, int -> acc + int }
+
+      sum shouldBe a * 6
+    }
+  }
+
+  // Note: this takes 50-60 sec to run
```
nomisRev, line 43: We should use KotlinX Coroutines Test here: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-blocking-test.html

That should make this test instant, whilst it should also stabalise any flakyness that might occur by doing time measurements on the JVM.

See: https://github.com/arrow-kt/arrow-fx/blob/master/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/BracketCaseTest.kt#L21

nomisRev (commented):
Looks good. Only one concern with the long running test, and the missing `counter` definition in the doc snippet.

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/FlowTest.kt
LordRaydenMK: Didn't know this existed. Today I learned :smile:

nomisRev (approved):
Awesome new operator 👏 👏 👏 Thanks @LordRaydenMK!

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/FlowTest.kt
```diff
@@ -0,0 +1,69 @@
...
+      val sum = flow {
+        emit(a)
+        if (++counter <= 5) throw RuntimeException("Bang!")
+      }.retry(Schedule.recurs(n))
+        .reduce { acc, int -> acc + int }
+
+      sum shouldBe a * 6
+    }
+  }
+
```
aballano, line 43: I would add an additional test to make sure we don't retry on CancellationException

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/flow.kt
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param schedule - the [Schedule] used for retrying the collection of the flow
+ */
+fun <A, B> Flow<A>.retry(schedule: Schedule<Throwable, B>): Flow<A> = flow {
+  (schedule as Schedule.ScheduleImpl<Any?, Throwable, B>)
+  var dec: Schedule.Decision<Any?, B>
+  var state: Any? = schedule.initialState()
+
+  val retryWhen = retryWhen { cause, _ ->
+    dec = schedule.update(cause, state)
```
aballano, line 44: Shouldn't we check for a possible cancellation exception here and rethrow?

aballano (commented):
Looks quite good, only a question regarding cancellation

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/flow.kt
LordRaydenMK: That should be handled by `retryWhen` 

https://github.com/Kotlin/kotlinx.coroutines/blob/22087ef122b5ac677fd14941a18b6aefc57393d2/kotlinx-coroutines-core/common/src/flow/operators/Errors.kt#L207

aballano (approved):

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/flow.kt
aballano: You're right, thank you!

arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/FlowTest.kt
LordRaydenMK: Since that is handled by `retryWhen` do we really need a test?

PR #383, opened 2021-01-20
Deprecate Reactor in favor of KotlinX Coroutines Reactor

nomisRev:
Since Arrow Fx Coroutines is going to be the new way forward over `IO` or polymorphic effect typeclasses the Arrow Fx Reactor module is deprecated in favor of using Arrow Fx Coroutines + KotlinX Coroutines Reactor.

```kotlin
import kotlinx.coroutines.reactor.mono
import kotlinx.coroutines.reactor.asCoroutineDispatcher
import reactor.core.scheduler.Scheduler

val mono: Mono<Pair<String, String>> = mono {
  parMapN(Schedulers.computation().asCoroutineDispatcher(),
    { Thread.currentThread().name },
    { Thread.currentThread().name },
    ::Pair
  )
}
```


aballano (approved):

raulraja (approved):
thanks @nomisRev !

franciscodr (approved):

i-walker (approved):

PR #384, opened 2021-01-20
Deprecate RxJava

nomisRev:
Since Arrow Fx Coroutines is going to be the new way forward over `IO` or polymorphic effect typeclasses the Arrow Fx RxJava module is deprecated in favor of using Arrow Fx Coroutines + KotlinX Coroutines RxJava.

```kotlin
import kotlinx.coroutines.rx2.mono
import kotlinx.coroutines.rx2.asCoroutineDispatcher
import io.reactive.Scheduler

val single: Single<Pair<String, String>> = single {
  parMapN(Schedulers.computation().asCoroutineDispatcher(),
    { Thread.currentThread().name },
    { Thread.currentThread().name },
    ::Pair
  )
}
```


aballano (approved):

franciscodr (approved):
Thanks @nomisRev!

arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/FlowableK.kt
```diff
@@ -245,6 +275,7 @@ data class FlowableK<out A>(val flowable: Flowable<out A>) : FlowableKOf<A> {
     fun <A> cancelableF(fa: ((Either<Throwable, A>) -> Unit) -> FlowableKOf<CancelToken<ForFlowableK>>, mode: BackpressureStrategy = BackpressureStrategy.BUFFER): FlowableK<A> =
```
danimontoya, line 275: The deprecation message should be change for `DeprecateRxJava` ?

arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/FlowableK.kt
```diff
@@ -228,6 +257,7 @@ data class FlowableK<out A>(val flowable: Flowable<out A>) : FlowableKOf<A> {
     fun <A> cancelable(fa: ((Either<Throwable, A>) -> Unit) -> CancelToken<ForFlowableK>, mode: BackpressureStrategy = BackpressureStrategy.BUFFER): FlowableK<A> =
```
danimontoya, line 257: The deprecation message should be change for `DeprecateRxJava` ?

i-walker (approved):

danimontoya (approved):
Some functions were already deprecated because of renaming.. maybe those deprecation messages need to be changed to `DeprecateRxJava` ?

PR #386, opened 2021-01-20
Sv update arrow fx docs

nomisRev:
Removes deprecated references from the docs, and adds a small section on the rationale of replacing `IO` with `suspend`.
I also wanted to tackle https://arrow-kt.io/docs/fx/polymorphism/ but couldn't find it in the `arrow-docs` project... 🤔

The `io` page still needs to be added to the website for `0.12.0`, and I'm not sure if we want to keep it later.
It might be useful as page for people used to working with `IO`.


arrow-docs/docs/io/README.md
```diff
@@ -12,3 +12,206 @@ Arrow has adopted suspend as system to model monadic computations and offers the
...
+Important here is that a `Throwable` that occurs in an async thread is safely captured in the `IO` as well and can be recovered from at any point.
+
+`suspend` always results in `Result<A>` which is isomorphic to `Either<Throwable, A>`, and it can be used to offer the same safety guarantees as `IO`.
+So the `suspend` API can also always return any exception safely to the user, and it can be recovered from at any point in the code.
+In contrast to `IO`, we can only find `startCoroutine` in the standard library and it's isomorphic to `unsafeRunAsync`.
+Instead of `f: (Either<Throwable, A>) -> Unit` you provide `f: (Result<A>) -> Unit` to run the `suspend () -> A` program.
+
+So `suspend () -> A` offers us the exact same guarantees as `IO<A>`.
+
+Both also forces us to go through certain set of methods to execute/run the effectful program.
```
i-walker, line 74: Nit: 
can we mention which methods or a link to some docs? Or is this rather a narrative style to lead to the next section or topics which are discussed, if it's the latter, maybe we can make this a bit more clear.

arrow-docs/docs/io/README.md
```diff
@@ -12,3 +12,206 @@ Arrow has adopted suspend as system to model monadic computations and offers the
...
+    }
+  }
+```
+
+So we can see that for this simple program we allocated 6 `IO` cases of the `IO` `sealed class`.
+This is necessary so when you `unsafeRun` the `IO` program, it can be inspected to find out which branch of code needs to run. 
+
+In contrast to this `suspend` can simply be wired by the Kotlin compiler, which eliminates the need for this `sealed class` and allocations.
+The Kotlin compiler rewrites the suspend program to a super fast runtime which uses a switch table and mutable state machine to run the `suspend` program.
+Furthermore, the Kotlin compiler applies a bunch of optimisations that improve the speed of `suspend` even further and certains JVM .
```
i-walker, line 217: nit
```suggestion
Furthermore, the Kotlin compiler applies a bunch of optimisations that improve the speed of `suspend` even further and certains JVM.
```

i-walker:
👏  Thanks simon,
I think what we should also mention is that by moving to suspended, users and library authors alike can easily integrate Arrow Fx in other effectfull code and libraries, without the integration modules that were previously needed.

arrow-docs/docs/io/README.md
raulraja, line 217:
```suggestion
Furthermore, the Kotlin compiler applies other optimisations focused on the speed of `suspend` and memory usage of suspension making it suitable for hotspots and places where otherwise allocations or ADT based data types are not an option.
```

raulraja (approved):
Thanks, @nomisRev .
I added some suggestions, feel free to apply whatever you think it makes sense if any.

arrow-docs/docs/io/README.md
```diff
@@ -12,3 +12,206 @@ Arrow has adopted suspend as system to model monadic computations and offers the
...
+import arrow.core.computations.either
+
+suspend fun suspendProgram(): Either<PersistenceError, ProcessedUser> =
+  either {
+    val user = !fetchUser()
+    val processed = !user.process()
+    processed
+  }
+```
+
```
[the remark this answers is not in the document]
nomisRev:
```suggestion
The above two examples demonstrate how much simpler `suspend` is over its `IO` counterpart and how the `either` computation block allows us to bind values of `Either` to extract their right side all while inside `suspend`.

Arrow Fx allows to intermix effects in suspension. What otherwise would have required the `EitherT` transformer over `IO` now it can just be expressed by wrapping in `either` instead
```

arrow-docs/docs/io/README.md
nomisRev:
```suggestion
Furthermore, the Kotlin compiler applies a bunch of optimizations that improve the speed of `suspend` even further.
```

i-walker (approved):

i-walker:
Fixed the build, ready to be merged if you are @nomisRev 🙌

PR #388, opened 2021-01-25
Remove Deprecated `Id<A>` data type

danimontoya:


arrow-fx/src/test/kotlin/arrow/fx/ScheduleTest.kt
```diff
@@ -191,49 +78,6 @@ class ScheduleTest : ArrowFxSpec() {
...
-        val sum = res.fold(0L) { acc, v ->
-          acc + v.delay.amount
-        }
-        val fibSum = fibs(i.toLong()).drop(1).take(res.size).sum()
-
-        res.forAll { it.cont } && sum == fibSum
-      }
-    }
-
-    "Schedule.linear()" {
```
i-walker, line 215: Are those to be replaced, with something else, or is this not required anymore? 
cc @nomisRev

nomisRev: Arrow Fx is deprecated and will be removed in 0.13.0 so this doesn't need to be replaced.

nomisRev (approved):
Thanks @danimontoya

arrow-fx/src/test/kotlin/arrow/fx/ScheduleTest.kt
danimontoya: thanks I was asking the same myself 👍

i-walker (approved):

PR #366, opened 2020-12-30
Deprecate arrow.fx.coroutines.Duration

i-walker:
## Issue
Initial deprecation cycle in favor of kotlin.time.Duration

links to: 
#18 

## Status
**READY**

## Description
This PR goes along with the changes introduced in #360 , #359

PR #359, opened 2020-12-23
Move Schedule's Duration to kotlin.time.Duration

i-walker:
## Status
**IN DEVELOPMENT**

## Description
The Schedule tests run into Casting Error's in their `delay`implementations.  
<details closed>
<summary>- Schedule.fibonacci()</summary>
java.lang.ClassCastException: class kotlin.Pair cannot be cast to class java.lang.Number (kotlin.Pair is in unnamed module of loader 'app'; java.lang.Number is in module java.base of loader 'bootstrap')
	at arrow.fx.coroutines.ScheduleTestKt.calculateSchedule(ScheduleTest.kt:241)
	at arrow.fx.coroutines.ScheduleTest$1$16.invokeSuspend(ScheduleTest.kt:126)
	at arrow.fx.coroutines.ScheduleTest$1$16.invoke(ScheduleTest.kt)
	at io.kotest.core.internal.ExecutionsKt$executeWithBehaviours$2$1.invokeSuspend(executions.kt:13)
	at io.kotest.core.internal.ExecutionsKt$executeWithBehaviours$2$1.invoke(executions.kt)
	at io.kotest.core.internal.ExecutionsKt.wrapTestWithGlobalAssert(executions.kt:39)
	at io.kotest.core.internal.ExecutionsKt$executeWithBehaviours$2.invokeSuspend(executions.kt:12)
	at io.kotest.core.internal.ExecutionsKt$executeWithBehaviours$2.invoke(executions.kt)
	at io.kotest.core.internal.ExecutionsKt$wrapTestWithAssertionModeCheck$2.invokeSuspend(executions.kt:25)
	at io.kotest.core.internal.ExecutionsKt$wrapTestWithAssertionModeCheck$2.invoke(executions.kt)
	at io.kotest.core.internal.AssertionsCheckKt.executeWithAssertionsCheck(assertionsCheck.kt:25)
	at io.kotest.core.internal.ExecutionsKt.wrapTestWithAssertionModeCheck(executions.kt:24)
	at io.kotest.core.internal.ExecutionsKt.executeWithBehaviours(executions.kt:11)
	at io.kotest.core.internal.TestCaseExecutor$executeInScope$2.invokeSuspend(TestCaseExecutor.kt:258)
	at io.kotest.core.internal.TestCaseExecutor$executeInScope$2.invoke(TestCaseExecutor.kt)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:91)
[...]

Issue #18, opened 2019-11-14
["Request"] Calculations on `Duration.INFINITE` should not overflow


supersven:
## What version are you currently using?

Latest master

## What would you like to see?

Calculations like `times` and `plus` on `Duration.INFINITE` should not overflow. Because this issue is due to calculations based on `Long`, it is not limited to `Duration.INFINITE`, but here the case is most obvious.

Additionally, it would be nice if `nanoseconds` would always be "valid" for the given `amount`. I.e. currently a large `amount` leads to `Long.MAX_VALUE` in `nanoseconds`, where a larger `nanoseconds` value would be right.

## Example

Currently
```
    "INFINITE plus one should be INFINITE" {
        val d = Duration(1L, TimeUnit.NANOSECONDS)
        Duration.INFINITE.plus(d).nanoseconds shouldBe  Duration.INFINITE.nanoseconds
    }
```
results in
```
expected: 9223372036854775807L but was: -9223372036854775808L
org.opentest4j.AssertionFailedError: expected: 9223372036854775807L but was: -9223372036854775808L
	at arrow.fx.data.DurationTest$3.invokeSuspend(DurationTest.kt:32)
	at arrow.fx.data.DurationTest$3.invoke(DurationTest.kt)
	at io.kotlintest.runner.jvm.TestCaseExecutor$executeTest$supervisorJob$1$invokeSuspend$$inlined$map$lambda$1.invokeSuspend(TestCaseExecutor.kt:130)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:241)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

nomisRev:
Closing this since we're deprecating our own `Duration`. If by the time of removal `kotlin.time.Duration` is still not stable we'll provide `Long` (milliseconds) based alternative methods.

PR #389, opened 2021-01-25
Improve Schedule for 1.0.0

nomisRev:
## Issue
 - Reported by @aballano: `arrow.fx.coroutines.repeat` conflicts with `kotlin.repeat` which IDEA always auto-suggests and prefers over `arrow.fx.coroutines.repeat` since `kotlin` is a special case.
 - Deprecation of Arrow Fx Coroutines Duration in favor of the upcoming kotlin.time.Duration

## Status
READY

## Description

Since `arrow.fx.coroutines.repeat` conflicts with `kotlin.repeat` which IDEA always auto-suggests and prefers over `arrow.fx.coroutines.repeat` since `kotlin` is a special case, we need an alternative API for `repeat` (and `retry`) so this is now defined as a method on `Schedule` itself.
Since `retry`'s Schedule has as input type of `Throwable` it's defined as a extension function on `Schedule` instead of a top-level function that takes a `Schedule`.

Since Arrow Fx Coroutines Duration is deprecated in favor of Kotlin Time's Duration we need to provide alternative APIs to deprecate towards. This PR adds these alternative methods, and the required deprecation messages.

## Related PRs
* #366

### DIFF
--- a/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
+++ b/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
@@ -7,12 +7,16 @@ import arrow.core.left
 import arrow.core.right
 import arrow.fx.coroutines.Schedule.ScheduleImpl
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.ensureActive
+import kotlin.coroutines.coroutineContext
 import kotlin.math.max
 import kotlin.math.min
 import kotlin.math.pow
 import kotlin.math.roundToInt
-import kotlin.math.roundToLong
 import kotlin.random.Random
+import kotlin.time.ExperimentalTime
+import kotlin.time.nanoseconds
+import arrow.fx.coroutines.nanoseconds as oldNanoseconds
 
 /**
  * # Retrying and repeating effects
@@ -74,7 +78,7 @@ import kotlin.random.Random
  * suspend fun main(): Unit {
  *   var counter = 0
  *   //sampleStart
- *   val res = repeat(Schedule.recurs(3)) {
+ *   val res = (Schedule.recurs(3)).repeat {
  *     println("Run: ${counter++}")
  *   }
  *   //sampleEnd
@@ -92,11 +96,11 @@ import kotlin.random.Random
  * suspend fun main(): Unit {
  *   var counter = 0
  *   //sampleStart
- *   val res = repeat(Schedule.unit<Unit>() zipLeft Schedule.recurs(3)) {
+ *   val res = (Schedule.unit<Unit>() zipLeft Schedule.recurs(3)).repeat {
  *     println("Run: ${counter++}")
  *   }
  *   // equal to
- *   val res2 = repeat(Schedule.recurs<Unit>(3) zipRight Schedule.unit()) {
+ *   val res2 = (Schedule.recurs<Unit>(3) zipRight Schedule.unit()).repeat {
  *     println("Run: ${counter++}")
  *   }
  *   //sampleEnd
@@ -113,11 +117,11 @@ import kotlin.random.Random
  * suspend fun main(): Unit {
  *   var counter = 0
  *   //sampleStart
- *   val res = repeat(Schedule.identity<Int>() zipLeft Schedule.recurs(3)) {
+ *   val res = (Schedule.identity<Int>() zipLeft Schedule.recurs(3)).repeat {
  *     println("Run: ${counter++}"); counter
  *   }
  *   // equal to
- *   val res2 = repeat(Schedule.recurs<Int>(3) zipRight Schedule.identity<Int>()) {
+ *   val res2 = (Schedule.recurs<Int>(3) zipRight Schedule.identity<Int>()).repeat {
  *     println("Run: ${counter++}"); counter
  *   }
  *   //sampleEnd
@@ -134,12 +138,12 @@ import kotlin.random.Random
  * suspend fun main(): Unit {
  *   var counter = 0
  *   //sampleStart
- *   val res = repeat(Schedule.collect<Int>() zipLeft Schedule.recurs(3)) {
+ *   val res = (Schedule.collect<Int>() zipLeft Schedule.recurs(3)).repeat {
  *     println("Run: ${counter++}")
  *     counter
  *   }
  *   // equal to
- *   val res2 = repeat(Schedule.recurs<Int>(3) zipRight Schedule.collect<Int>()) {
+ *   val res2 = (Schedule.recurs<Int>(3) zipRight Schedule.collect<Int>()).repeat {
  *     println("Run: ${counter++}")
  *     counter
  *   }
@@ -159,7 +163,7 @@ import kotlin.random.Random
  * suspend fun main(): Unit {
  *   var counter = 0
  *   //sampleStart
- *   val res = repeat(Schedule.doWhile<Int>{ it <= 3 }) {
+ *   val res = Schedule.doWhile<Int>{ it <= 3 }.repeat {
  *     println("Run: ${counter++}"); counter
  *   }
  *   //sampleEnd
@@ -179,10 +183,26 @@ import kotlin.random.Random
  */
 sealed class Schedule<Input, Output> {
 
+  abstract suspend fun <C> repeatOrElseEither(fa: suspend () -> Input, orElse: suspend (Throwable, Output?) -> C): Either<C, Output>
+
+  /**
+   * Runs this effect once and, if it succeeded, decide using the provided policy if the effect should be repeated and if so, with how much delay.
+   * Returns the last output from the policy or raises an error if a repeat failed.
+   */
+  suspend fun repeat(fa: suspend () -> Input): Output =
+    repeatOrElse(fa) { e, _ -> throw e }
+
+  /**
+   * Runs this effect once and, if it succeeded, decide using the provided policy if the effect should be repeated and if so, with how much delay.
+   * Also offers a function to handle errors if they are encountered during repetition.
+   */
+  suspend fun repeatOrElse(fa: suspend () -> Input, orElse: suspend (Throwable, Output?) -> Output): Output =
+    repeatOrElseEither(fa, orElse).fold(::identity, ::identity)
+
   /**
    * Changes the output of a schedule. Does not alter the decision of the schedule.
    */
-  abstract fun <B> map(f: (Output) -> B): Schedule<Input, B>
+  abstract fun <B> map(f: (output: Output) -> B): Schedule<Input, B>
 
   /**
    * Changes the input of the schedule. May alter a schedule's decision if it is based on input.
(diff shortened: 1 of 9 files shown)


arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -74,7 +78,7 @@ import kotlin.random.Random
  * suspend fun main(): Unit {
  *   var counter = 0
  *   //sampleStart
- *   val res = repeat(Schedule.recurs(3)) {
+ *   val res = (Schedule.recurs(3)).repeat {
```
aballano, line 81:
```suggestion
 *   val res = Schedule.recurs(3).repeat {
```

i-walker (approved):
🙌 Thanks @nomisRev for this !! 
With this, I'll reassess if #359 is still relevant

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -192,21 +212,48 @@ sealed class Schedule<Input, Output> {
...
+  fun <A : Input, B> combineWith(
     other: Schedule<A, B>,
     f: (Boolean, Boolean) -> Boolean,
     g: (Duration, Duration) -> Duration
-  ): Schedule<A, Pair<Output, B>>
+  ): Schedule<A, Pair<Output, B>> =
+    combineNanos(other, f, { a, b -> g(a.toLong().oldNanoseconds, b.toLong().oldNanoseconds).nanoseconds.toDouble() }, ::Pair)
+
+  /**
+   * Combines with another schedule by combining the result and the delay of the [Decision] with the functions [f], [g] and a [zip] function
```
aballano, line 237:
```suggestion
   * Combines with another schedule by combining the result and the delay of the [Decision] with the [zipContinue], [zipDuration] and a [zip] functions
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -192,21 +212,48 @@ sealed class Schedule<Input, Output> {
...
+   */
+  @ExperimentalTime
+  fun <A : Input, B, C> combine(
+    other: Schedule<A, B>,
+    zipContinue: (cont: Boolean, otherCont: Boolean) -> Boolean,
+    zipDuration: (duration: kotlin.time.Duration, otherDuration: kotlin.time.Duration) -> kotlin.time.Duration,
```
aballano, line 243: nit: you could import this with an alias like `kotlinDuration` or something like that

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -192,21 +212,48 @@ sealed class Schedule<Input, Output> {
...
+    zip: (Output, B) -> C
+  ): Schedule<A, C> =
+    combineNanos(other, zipContinue, { a, b -> zipDuration(a.nanoseconds, b.nanoseconds).inNanoseconds }, zip)
+
+  /**
+   * Combines with another schedule by combining the result and the delay of the [Decision] with the functions [f], [g] and a [zip] function
```
aballano, line 249:
```suggestion
   * Combines with another schedule by combining the result and the delay of the [Decision] with the [zipContinue], [zipDuration] and a [zip] functions
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -192,21 +212,48 @@ sealed class Schedule<Input, Output> {
...
    * Inverts the decision of a schedule.
    */
   abstract operator fun not(): Schedule<Input, Output>
 
-  /**
-   * Combines with another schedule by combining the result and the delay of the [Decision] with the functions [f] and [g]
-   */
-  abstract fun <A : Input, B> combineWith(
+  @Deprecated(
+    "combineWith is deprecated in favor of combineNanos and combine using Double/kotlin.time.Duration instead of arrow.fx.coroutines.Duration.",
```
aballano, line 223: wouldn't be easier replacing with the kotlin.Duration version?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -218,53 +265,81 @@ sealed class Schedule<Input, Output> {
    */
   abstract infix fun <A : Input, B> andThen(other: Schedule<A, B>): Schedule<A, Either<Output, B>>
 
+  @Deprecated(
+    "modifyDelay will be replaced by modify and modifyNanos which use kotlin.time.Duration instead of arrow.fx.coroutines.Duration",
+    ReplaceWith("modifyNanos { output, l -> f(output, l.toLong().nanoseconds).nanoseconds.toDouble() }", "arrow.fx.coroutines.nanoseconds")
```
aballano, line 270: `f(output, l.toLong().nanoseconds).nanoseconds.toDouble()` yep, I definitely prefer the kotlin.Duration version 😄

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -218,53 +265,81 @@ sealed class Schedule<Input, Output> {
...
-  abstract fun logInput(f: suspend (Input) -> Unit): Schedule<Input, Output>
+  abstract fun logInput(f: suspend (input: Input) -> Unit): Schedule<Input, Output>
 
   /**
    * Runs an effectful handler on every output. Does not alter the decision.
    */
-  abstract fun logOutput(f: suspend (Output) -> Unit): Schedule<Input, Output>
+  abstract fun logOutput(f: suspend (output: Output) -> Unit): Schedule<Input, Output>
+
+  @Deprecated("foldM is renamed to fold. Use fold instead.", ReplaceWith("foldLazy(initial, f)"))
```
aballano, line 295:
```suggestion
  @Deprecated("foldM is renamed to foldLazy.", ReplaceWith("foldLazy(initial, f)"))
```

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -297,55 +372,59 @@ sealed class Schedule<Input, Output> {
...
    */
   infix fun <A : Input, B> zipLeft(other: Schedule<A, B>): Schedule<A, Output> =
-    (this and other).map { it.first }
+    (this and other).map(Pair<Output, B>::first)
 
   /**
    * Adjusts the delay of a schedule's [Decision].
    */
-  fun delayed(f: suspend (Duration) -> Duration): Schedule<Input, Output> =
+  @Deprecated("TODO")
```
aballano, line 398: leftover?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -519,41 +625,100 @@ sealed class Schedule<Input, Output> {
   /**
    * A single decision. Contains the decision to continue, the delay, the new state and the (lazy) result of a Schedule.
    */
-  data class Decision<out A, out B>(val cont: Boolean, val delay: Duration, val state: A, val finish: Eval<B>) {
+  data class Decision<out A, out B>(val cont: Boolean, val delayInNanos: Double, val state: A, val finish: Eval<B>) {
+
+    constructor(cont: Boolean, delay: Duration, state: A, finish: Eval<B>) : this(cont, delay.nanoseconds.toDouble(), state, finish)
+
+    @Deprecated(
+      "Arrow Fx Duration is deprecated use Double or kotlin.time.Duration",
```
aballano, line 633: I think the message should indicate to use an alternative property, not a type

```suggestion
      "Arrow Fx Duration is deprecated, use duration or delayInNanos",
```

or alternatively follow the same structure as previous deprecated messages

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
aballano, line 223: Also, I'd suggest aligning the deprecated messages if possible:

- "is deprecated in favor of"
- "will be replaced by"
- "is renamed to"

aballano (commented):
First quick review with some nits and wording alternatives, will take another look later today

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
nomisRev: Yes it absolutely would but `kotlin.time.Duration` is still `@Experimental` and this would make a large portion of the `Schedule` API `@Experimental` as well while in it's not experimental at all.

nomisRev:
Thanks for the great review @aballano!

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
aballano: Makes sense, but I still believe that replacing Duration with nanos is not a good idea. I'd personally rather do a migration to something experimental but do it once than do it to something flaky (like nanos with Double) to then migrate again to Duration.

In summary: 

* Promoting nanos API -> we're promoting a flaky API (I believe it's really easy switching millis with nanos and break something) + users will likely need to do an additional migration to Duration
* Promoting Duration API -> we promote what's in Kotlin + there's no extra migration needed + users can still choose the nanos API if they don't want to deal with experimental APIs

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -703,15 +968,61 @@ sealed class Schedule<Input, Output> {
     /**
      * Creates a Schedule that continues with a fixed delay.
      */
+    @Deprecated(
+      DeprecatedDurationAPI,
+      ReplaceWith(
+        "Schedule.forever<A>().delayedNanos { d -> d + interval.nanoseconds }",
+        "arrow.fx.coroutines.Schedule"
```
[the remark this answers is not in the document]
nomisRev: There is no `toDouble` on `arrow.fx.coroutines.Duration` so that would have to still rely on `@ExperimentalTime` with `kotlin.time.nanoseconds`. :(

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
nomisRev: >  I still believe that replacing Duration with nanos is not a good idea

We need to use `Double` nanos in the implementation or the implemenation itself becomes `@ExperimentalTime` which means that *all* APIs would become `@ExperimentalTime` also when they don't rely on `Duration` in their API anywhere.

I don't want to promote `Double` as nanos in the public APIs, and I'd rather also not expose it but at this point we don't really have a choose. We could also use `Long` and `milliseconds` in the public API, I guess that is more common and similar to what KotlinX Coroutines does with `delay`.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -703,15 +968,61 @@ sealed class Schedule<Input, Output> {
...
+      forever<A>().delayedNanos { d -> d + interval.inNanoseconds }
 
     /**
      * Creates a Schedule that continues with increasing delay by adding the last two delays.
      */
+    @Deprecated(
+      DeprecatedDurationAPI,
+      ReplaceWith(
+        "Schedule.delayed(Schedule.forever<A>().map { base * it })",
+        "arrow.fx.coroutines.Schedule"
```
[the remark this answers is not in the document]
nomisRev: Idem

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -720,19 +1031,57 @@ sealed class Schedule<Input, Output> {
...
      * Delays can be calculated as [base] * factor ^ n where n is the number of executions.
      */
-    fun <A> exponential(base: Duration, factor: Double = 2.0): Schedule<A, Duration> =
-      delayed(
-        forever<A>().map { base * factor.pow(it).roundToInt() }
+    @Deprecated(
+      DeprecatedDurationAPI,
+      ReplaceWith(
+        "Schedule.delayed(Schedule.forever<A>().map { base * factor.pow(it).roundToInt() })",
+        "arrow.fx.coroutines.Schedule"
```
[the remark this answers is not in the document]
nomisRev: Idem.

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -676,13 +921,33 @@ sealed class Schedule<Input, Output> {
...
+      (forever<A>() as ScheduleImpl<Int, A, Int>).reconsider { _: A, decision ->
+        Decision(
+          cont = decision.cont,
+          delayInNanos = decision.delayInNanos,
+          state = decision.state,
+          finish = Eval.now(decision.delayInNanos)
+        )
+      }
+
+    fun <A> duration(): Schedule<A, kotlin.time.Duration> =
```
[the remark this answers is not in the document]
nomisRev: We cannot name both `delay` as this conflicts their signatures.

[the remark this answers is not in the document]
aballano: but we can use `@JvmName` for that, no?

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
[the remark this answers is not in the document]
aballano: nevermind, I though the `nanoseconds` was an ext-fun, but is not, so this is not necessary to do `Schedule.spaced<A>(interval.nanoseconds.toDouble())`

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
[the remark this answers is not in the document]
aballano: Fixed, removed the extra import

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
i-walker: Not to mention that we also need a notion of infinity https://github.com/arrow-kt/arrow-fx/issues/18

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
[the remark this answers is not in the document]
nomisRev: No, the `@JvmName` trick only works for things that conflict in the JVM but not in Kotlin. In this case, it's also a conflict in Kotlin.

aballano (approved):
Done reviewing, feel free to tackle the comments that make sense (and ignore those that do not :D)

arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/Schedule.kt
```diff
@@ -703,15 +968,61 @@ sealed class Schedule<Input, Output> {
     /**
      * Creates a Schedule that continues with a fixed delay.
      */
+    @Deprecated(
+      DeprecatedDurationAPI,
+      ReplaceWith(
+        "Schedule.forever<A>().delayedNanos { d -> d + interval.nanoseconds }",
```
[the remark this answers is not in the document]
nomisRev: These suggested changes are incorrect. They return an incorrect Schedule.

nomisRev:
@aballano applied review suggestions. The only thing I can think of to improve the `nanos` situation is to make the API use `milliseconds` instead which I guess is more common in public APIs. `kotlin.time.*` should just go stable 😫

PR #397, opened 2021-02-15
This commit enables indentation in .editorconfig file so that indenta…

Daniel-Jacob:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,4 +3,4 @@
 indent_size=2
 insert_final_newline=true
 max_line_length=off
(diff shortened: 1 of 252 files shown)


rachelcarmena (approved):
Thanks @Daniel-Jacob !! I couldn't review 240 files manually so I made it with `tr` and `diff` commands and I can confirm that the PR just includes format changes. A new repository was added in `build.gradle` that is going to disappear very soon so don't worry, it will be fixed in the main repository. Thanks!!

i-walker (approved):

